PR #5, opened 2021-02-01
Expose send_message_to_group

boxdot:


src/manager.rs
```diff
@@ -455,6 +456,50 @@ where
...
+
+    pub async fn send_message_to_group(
+        &self,
+        recipients: impl IntoIterator<Item = String>,
+        message: DataMessage,
+        timestamp: u64,
+    ) -> Result<(), Error> {
+        let mut sender = self.get_sender()?;
+
+        let recipients: Vec<_> = recipients
```
gferon, line 482: I guess I should use `impl IntoIterator<Item = String>` instead in the lib?

PR #8, opened 2021-02-07
Sled tests + group msg CLI command 

boxdot:
Based on #7


src/main.rs
```diff
@@ -106,7 +121,7 @@ async fn main() -> anyhow::Result<()> {
 
     match args.subcommand {
         Subcommand::Config { command } => match command {
-            ConfigSubcommand::Print => println!("{:?}", manager.config_store),
+            ConfigSubcommand::Print => println!("{:?}", manager.config_store.keys()),
```
gferon, line 124: very good! 🔒

PR #12, opened 2021-04-24
Add support for fetching contacts from sync message

gferon:


gferon:
@boxdot you might want to have a look

PR #16, opened 2021-04-28
Fix panic when receiving messages if the device is not registered.

boxdot:
Also expose `is_registered` method and remove optional credentials
state.


gferon:
Excellent, thanks!

PR #17, opened 2021-04-29
Use a single ConfigStore param in Manager.

boxdot:
The ConfigStore trait already requires all other trais being
implmented, so there is no need to store them.

This also makes the random generator parameter default.


gferon (approved):
Amazing, didn't think of the trick with `C, C, C, C` which is why I couldn't make it work and also didn't know `R = ConcreteType` for template parameters, thanks!

PR #19, opened 2021-04-30
Cache push service and factor out other factory methods.

boxdot:
Depends on https://github.com/Michael-F-Bryan/libsignal-service-rs/pull/93

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -10,7 +10,7 @@ use qrcode::QrCode;
 use rand::{distributions::Alphanumeric, CryptoRng, Rng, RngCore};
 
 use libsignal_service::{
-    cipher::ServiceCipher,
+    cipher,
     configuration::{ServiceConfiguration, SignalServers, SignalingKey},
     content::{ContentBody, DataMessage, Metadata},
     groups_v2::{GroupsManager, InMemoryCredentialsCache},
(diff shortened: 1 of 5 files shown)


src/manager.rs
```diff
@@ -33,13 +33,36 @@ use libsignal_service::{
 
 use libsignal_service_hyper::push_service::HyperPushService;
 
+use crate::cache::CacheCell;
 use crate::{config::ConfigStore, Error};
 
+type ServiceCipher<C, R> = cipher::ServiceCipher<C, C, C, C, R>;
+type MessageSender<C, R> =
```
gferon, line 40: awesome!

src/manager.rs
```diff
@@ -147,14 +181,9 @@ where
         let rng = rand::rngs::OsRng::default();
         let password: String = rng.sample_iter(&Alphanumeric).take(24).collect();
 
-        let cfg: ServiceConfiguration = signal_servers.into();
+        let mut push_service = self.push_service()?;
```
gferon, line 184: this is not gonna work, because we aren't registered yet: you should initialize a `PushService` with the provided `signal_servers` instead of getting them from the state.

boxdot: ProvisioningManager now override the auth. This was merged.

gferon: yes, but `fn push_service()` now inspects the state and if it's `State::New` returns an error, which can't work for registration / linking.

boxdot: Ah, this is wrong then. The function below needs to be fixed.

gferon: I have some improvement working on top of your PR that's also related, I'll add one commit on top that we can have a look at together.

boxdot: Perfect. Waiting for it.

gferon:
I'm merging this PR as-is, even though it breaks linking and registration, because I'll push a small refactor of `State` on top of it that restores the functionality.

PR #20, opened 2021-05-02
Store contacts from synchronization

gferon:
Solves #22

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,9 +32,9 @@ tokio = { version = "1.0", features = [ "macros", "rt-multi-thread" ] }
 quickcheck = "1.0.3"
 quickcheck_async = "0.1"
 
-#[patch."https://github.com/Michael-F-Bryan/libsignal-service-rs.git"]
-#libsignal-service = { path = "../libsignal-service-rs/libsignal-service" }
-#libsignal-service-hyper = { path = "../libsignal-service-rs/libsignal-service-hyper" }
+[patch."https://github.com/gferon/libsignal-service-rs.git"]
+libsignal-service = { path = "../libsignal-service-rs/libsignal-service" }
+libsignal-service-hyper = { path = "../libsignal-service-rs/libsignal-service-hyper" }
 
 [features]
 quirks = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,9 +32,9 @@ tokio = { version = "1.0", features = [ "macros", "rt-multi-thread" ] }
 quickcheck = "1.0.3"
 quickcheck_async = "0.1"
 
-[patch."https://github.com/gferon/libsignal-service-rs.git"]
-libsignal-service = { path = "../libsignal-service-rs/libsignal-service" }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -32,9 +32,9 @@ tokio = { version = "1.0", features = [ "macros", "rt-multi-thread" ] }
 quickcheck = "1.0.3"
 quickcheck_async = "0.1"
 
-#[patch."https://github.com/Michael-F-Bryan/libsignal-service-rs.git"]
-#libsignal-service = { path = "../libsignal-service-rs/libsignal-service" }
-#libsignal-service-hyper = { path = "../libsignal-service-rs/libsignal-service-hyper" }
+[patch."https://github.com/gferon/libsignal-service-rs.git"]
+libsignal-service = { path = "../libsignal-service-rs/libsignal-service" }
+libsignal-service-hyper = { path = "../libsignal-service-rs/libsignal-service-hyper" }
```
boxdot, line 37: I think this broke the CI.

src/config/sled.rs
```diff
@@ -170,6 +173,28 @@ impl ConfigStore for SledConfigStore {
     }
 }
 
+impl ContactsStore for SledConfigStore {
+    fn save_contacts(&mut self, contacts: &[Contact]) -> Result<(), Error> {
+        // TODO: we probably want to store the contacts in a map of UUID > data
```
boxdot, line 178: `Uuid` is `Ord`, so I guess you could return a sorted Vec and let users use the binary search. This would require Ord on `ServiceAddress` though.

src/config/sled.rs
```diff
@@ -170,6 +173,28 @@ impl ConfigStore for SledConfigStore {
...
+            .insert("contacts", serde_json::to_vec(contacts)?)?;
+        trace!("saved contacts");
+        Ok(())
+    }
+
+    fn contacts(&self) -> Result<Vec<Contact>, Error> {
+        self.db
+            .read()
+            .expect("poisoned mutex")
+            .open_tree("contacts")?
```
boxdot, line 192: Just a not: I was thinking that probably we should use a constant prefix here. In the future, I wanted to use the same sled db for storing other information. Need to be careful not to conflict.

src/manager.rs
```diff
@@ -519,6 +520,8 @@ where
         Ok(())
     }
 
+    /// Request that the primary device to encrypt & send all of its contacts as a message to ourselves
+    /// which can be then received, decrypted and stored in the message receiving loop
```
boxdot, line 524: I guess you could add a small comment about a post condition, that after this method is successfully done, the contacts are stored in the store.

boxdot (commented):
Amazing. Looking forward to use it in the client.

src/config/sled.rs
gferon: But [`PhoneNumber`](https://docs.rs/phonenumber/0.3.1+8.12.9/phonenumber/struct.PhoneNumber.html) sadly isn't

PR #37, opened 2021-12-21
cli example, send: Accept uuid instead of phone number

g-braeunlich:
Reason: signal dropped support for phone numbers at the https://textsecure-service.whispersystems.org/v1/messages/ endpoint.
Also currently, there is no phone number to uuid mapping in presage.
As a consequence, it is currently not possible to send messages with the cli example.
Interim solution: specify uuid instead of phone number.


gferon:
Sorry for reacting so late to this, thanks a lot!

PR #46, opened 2022-01-28
Show qrcode in terminal

boxdot:
Before a PNG image was generated and opened with default viewer.


gferon:
Nice! Could you maybe have this with a feature flag though? I often run this from a window that's not big enough 🥲

boxdot:
Huh, I wanted to add the feature flag. :) Should I still do it?

gferon:
Just realized it was a bit unreasonable to ask haha. Do it only if it's not too much to ask!

Issue #53, opened 2022-05-17
failed to select a version for the requirement `curve25519-dalek = "^2.0.0"`


EcoloSweet:
With this simple Cargo.toml
```toml
[package]
edition = "2021"
name = "project"
version = "0.1.0"

[dependencies]
presage = {git = "https://github.com/whisperfish/presage"}
```

when I run `cargo update` I have this error : 
```bash
error: failed to select a version for the requirement `curve25519-dalek = "^2.0.0"`
candidate versions found which didn't match: 3.2.1
location searched: Git repository https://github.com/signalapp/curve25519-dalek.git?branch=lizard2
required by package `zkgroup v0.7.3 (https://github.com/signalapp/zkgroup?tag=v0.7.3#197c382e)`
    ... which satisfies git dependency `zkgroup` of package `libsignal-service v0.1.0 (https://github.com/whisperfish/libsignal-service-rs?rev=efd4ea86f57520d99141bb9c1c4b38a2ac646d6a#efd4ea86)`
    ... which satisfies git dependency `libsignal-service` of package `presage v0.2.0 (https://github.com/whisperfish/presage#8fcfb651)`
    ... which satisfies git dependency `presage` of package `project v0.1.0 (/tmp/project)`
```

Thank you !

EcoloSweet:
Maybe @rubdos knows what happened here, seems related to https://github.com/signalapp/curve25519-dalek/pull/4

rubdos:
You need to add a patch section to your own project.  This is how Signal builds their own too.  @gferon, you might want to add that to docs.

Currently, it's probably this:

```toml
[patch.crates-io]
curve25519-dalek = { git = 'https://github.com/signalapp/curve25519-dalek', branch = '3.0.0-lizard2' }
```

When Presage's dependencies get bumped again, you'll need the `lizard2` branch, because of the PR you linked above:

```toml
[patch.crates-io]
curve25519-dalek = { git = 'https://github.com/signalapp/curve25519-dalek', branch = 'lizard2' }
```

EcoloSweet:
I'm afraid I have the same error, even with this "3.0.0-lizard2" patch :

```bash
error: failed to select a version for the requirement `curve25519-dalek = "^2.0.0"`
candidate versions found which didn't match: 3.2.1
location searched: Git repository https://github.com/signalapp/curve25519-dalek.git?branch=lizard2
[...]

PR #54, opened 2022-05-20
Update libsignal-service to fix curve25519-dalek cargo error

EcoloSweet:
Fixes https://github.com/whisperfish/presage/issues/53


EcoloSweet:
well, not that simple in fact

PR #57, opened 2022-06-02
Reexport States and uuid() function

EcoloSweet:
Since this commit [Rewrite Manager to use type states](https://github.com/whisperfish/presage/commit/55772b71b10c8b8dfb8f8e9e92f5339c4d4d7a57) , States are not exported anymore (usefull to be able to manipulate a manager outside of presage), and uuid() method was removed

thank you


gferon:
I actually did this change deliberately as I was thinking most of the data inside the state is either a secret or some private implementation detail. What kind of stage manipulation are you actually doing?

PR #59, opened 2022-06-25
More generic way of linking device

Schmiddiii:
Currently linking a device prints the corresponding QR code into the terminal. This is not really usefull for e.g. graphical clients. This PR makes the linking more generic by providing a callback function that is called with the URL to link with.

This has one disadvantage that error handling is a little bit worse as one cannot exactly identify the type of the error, I therefore just created a unparametrized `LinkError` that is given if the callback returns an error.

I also made the `qr2term` package a optional but default feature so that it can be disabled for graphical applications.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ serde = "1.0"
 serde_json = "1.0"
 thiserror = "1.0"
 sled = { version = "0.34", optional = true }
-qr2term = "0.2.2"
+qr2term = { version = "0.2.2", optional = true}
 
 [dev-dependencies]
 # for tests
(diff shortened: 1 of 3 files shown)


gferon:
Thanks, this makes sense! It's a little funny because this was like this initially but it got simplified early on when all users of the library were either TUI or CLI.

Since we're using asynchronous code, you probably want to use a https://docs.rs/tokio/latest/tokio/sync/oneshot/fn.channel.html instead of passing a function reference, though.

Schmiddiii:
Thanks for the feedback. I decided to use the Stream from futures as this is already a dependency of this crate and was already used many times. Note that oneshot channels will not work as the linking happens in a while loop and as oneshot senders are not clonable, this will not work.

The CI should also be fixed, just some small error in the `Cargo.toml`.

gferon:
Great, thanks! I have a bunch of other nitpicks that I'll take care of myself because I don't want to waste your time, namely:

1. I'm OK with breaking the API which means we shouldn't have two linking methods, I'd rather keep the API surface small.
2. I'll get rid of the `while` loop which was weird to do in the first place as both those enum variants come in a specific order anyways

PR #61, opened 2022-06-28
Export things needed to create custom ConfigStore

Schmiddiii:
Currently, implementing a custom `ConfigStore` outside of `presage` is not possible, mainly as `ContactsStore` and `StateStore` are not publicly accessible. Futhermore, some data in `Registered` needs to be public to be stored in the `StateStore`. This PR basically just makes all of these needed things public.

I also thought about maybe hiding the exports behind a feature, as most people will probably not need to create their own `ConfigStore`, but having the option is still needed as sometimes the built-in stores may not suffice.

Also note that the CI is currently failing, I did not change anything there so it will probably also fail in this PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,7 +7,7 @@ mod manager;
 pub use config::sled::SledConfigStore;
 pub use config::volatile::VolatileConfigStore;
 
-pub use config::ConfigStore;
+pub use config::{ConfigStore, ContactsStore, StateStore};
 pub use errors::Error;
 pub use manager::{Confirmation, Linking, Manager, Registered, Registration, RegistrationOptions};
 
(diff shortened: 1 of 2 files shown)


gferon:
Would it be possible to not have those fields public but just add a getter? I don't think it makes sense to expose the ability for users to swap the values in those structs.

gferon:
Thanks! I think I'd prefer to have getter that return an immutable reference to what needs to be accessed, as I don't see a use-case for modifying those values.

Schmiddiii:
Done. I have changed it back to `pub(crate)` instead of private so that the currently implemented stores (and #60) do not need to change.

gferon:
I'll merge it and fix the clippy warnings in the `main` branch.

PR #67, opened 2022-08-15
Message storage

### LINKED ISSUE #44: Add message storage
Move the message storage from https://github.com/boxdot/gurk-rs/ to presage, which should make it more useful. This will also make implementing encryption easier, since it'll be in a single place. Pinging @boxdot to fill in the details.

Schmiddiii:
This is a initial draft for message storage. The trait behind message storage is `MessageStore`, which is currently only has very basic functionality, e.g. storing and retrieving messages by sender and timestamp, listing all messages. This should definitely be expanded in the future with e.g. associating messages with channels/groups, listing messages from channels/groups. I don't think searching would be easily possible as sled (the main current implementation) is basically just a key-value store and not intended to be searchable (as far as I see).

The message serialization/deserialization is done using protobuf. I think the additional work done here (mainly the `InternalSerialization.proto` with function implementations) should be upstreamed to `libsignal-service-rs`

For receiving messages, there is a new function in the `Manager`, if the config store is also a messages store. This is only a thin wrapper around the normal receive function.
 
The example cli has also been updated to include use the new function for receiving and a argument for listing the current messages.

As already mentioned, this is still a early draft and only represents a idea on how this could be implemented. This idea should be discussed if this even is the right direction for this project.

This would close #44.

## Update

Associating messages with a contact/group is now implemented. I think the code is not as nice as it could be because `sled` does not seem to be made for such tasks, something like a real DB would be better suited. But it is working.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,33 @@
+use std::io::Result;
+use std::path::Path;
+
+fn main() -> Result<()> {
+    let protobuf = Path::new("src/protobuf").to_owned();
+
+    // Build script does not automagically rerun when a new protobuf file is added.
+    // Directories are checked against mtime, which is platform specific
+    println!("cargo:rerun-if-changed=src/protobuf");
+    // Adding src/proto.rs means an extra `include!` will trigger a rerun. This is on best-effort
+    // basis.
+    println!("cargo:rerun-if-changed=src/proto.rs");
+
+    let input: Vec<_> = protobuf
+        .read_dir()
+        .expect("protobuf directory")
+        .filter_map(|entry| {
+            let entry = entry.expect("readable protobuf directory");
+            let path = entry.path();
+            if Some("proto") == path.extension().and_then(std::ffi::OsStr::to_str) {
+                assert!(path.is_file());
+                println!("cargo:rerun-if-changed={}", path.to_str().unwrap());
+                Some(path)
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    prost_build::compile_protos(&input, &[protobuf])?;
+
+    Ok(())
+}
--- a/examples/cli.rs
+++ b/examples/cli.rs
@@ -94,16 +94,21 @@ enum Cmd {
     #[clap(about = "List contacts")]
     ListContacts,
     #[clap(about = "List messages")]
-    ListMessages,
-    #[clap(about = "List messages from contact")]
-    ListMessagesFromContact {
-        #[clap(long, short = 'u', help = "contact UUID")]
-        uuid: Uuid,
-    },
-    #[clap(about = "List messages")]
-    ListMessagesFromGroup {
-        #[clap(long, short = 'k', help = "Master Key of the V2 group (hex string)")]
-        master_key: String,
+    ListMessages {
+        #[clap(
+            long,
+            short = 'u',
+            help = "contact UUID",
+            conflicts_with = "master-key"
+        )]
+        uuid: Option<Uuid>,
+        #[clap(
+            long,
+            short = 'k',
+            help = "Master Key of the V2 group (hex string)",
+            conflicts_with = "uuid"
+        )]
+        master_key: Option<String>,
     },
     #[clap(about = "Find a contact in the embedded DB")]
     FindContact {
@@ -433,34 +438,22 @@ async fn run<C: ConfigStore + MessageStore>(
             let manager = Manager::load_registered(config_store)?;
             manager.request_contacts_sync().await?;
         }
-        Cmd::ListMessages => {
-            let mut msgs = config_store.messages()?;
-            msgs.sort_by_key(|m| m.metadata.timestamp);
-            for msg in msgs {
-                match msg.body {
-                    ContentBody::DataMessage(message)
-                    | ContentBody::SynchronizeMessage(SyncMessage {
-                        sent:
-                            Some(Sent {
-                                message: Some(message),
-                                ..
-                            }),
-                        ..
-                    }) => {
-                        println!(
-                            "{}: {}",
-                            msg.metadata.sender.e164_or_uuid(),
-                            message.body.unwrap_or_else(|| "".to_string())
-                        )
-                    }
-                    _ => {}
+        Cmd::ListMessages { master_key, uuid } => {
+            let thread = match (master_key, uuid) {
+                (Some(master_key), _) => {
+                    let master_key_bytes = hex::decode(master_key).expect("Master Key to be hex");
+                    Thread::Group(
+                        master_key_bytes
+                            .try_into()
+                            .expect("Master key to be 32 bytes"),
+                    )
                 }
-            }
-        }
-        Cmd::ListMessagesFromContact { uuid } => {
-            let thread = Thread::Contact(uuid);
-            let iter = config_store.messages_by_thread(&thread, None)?;
-            for msg in iter.take(10) {
+                (_, Some(uuid)) => Thread::Contact(uuid),
+                _ => panic!("At least one of master-key or uuid has to be given"),
+            };
+            let iter = config_store.messages(&thread, None)?;
+            let mut printed = 0;
+            for msg in iter {
                 match msg.body {
                     ContentBody::DataMessage(message)
                     | ContentBody::SynchronizeMessage(SyncMessage {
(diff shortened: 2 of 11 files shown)


gferon:
Awesome contribution 🥳 Thanks a lot @Schmiddiii! I'll try to give this a look ASAP and will get back to you.

gferon:
> something like a real DB would be better suited. But it is working.

That was my plan all along! I really wanted to spend some time extracting the storage backend from https://gitlab.com/whisperfish/whisperfish so we can share this, but I never got the courage to get started. Alternatively, we can look into replacing `sled` by something else.

Schmiddiii:
>That was my plan all along!  I really wanted to spend some time extracting the storage backend from https://gitlab.com/whisperfish/whisperfish so we can share this, but I never got the courage to get started.

Sounds interesting. Do you think that would be hard? One request I would have is also encrypted storage for that (I don't know if that is already implemented in Whisperfish). I already have encrypted storage for Flare, but it is pretty ugly, basically a clone of your `sled`-store that uses `encrypted-sled` instead.

gferon:
> Sounds interesting. Do you think that would be hard?

I don't think so, but rather tedious - I know @rubdos refactored some of the storage backend a little while ago and it might be worth looking at this with fresh eyes. Encryption is already supported and enabled with `sqlcipher` 🗝️.

Schmiddiii:
It is probably worth looking into that a little more before continuing with this PR.

build.rs
```diff
@@ -0,0 +1,33 @@
+use std::io::Result;
+use std::path::Path;
+
+fn main() -> Result<()> {
```
gferon, line 4: I think it's fine to have this here, instead of moving `InternalSerialization.proto` in `libsignal-service-rs`

examples/cli.rs
```diff
@@ -169,30 +190,32 @@ async fn send<C: ConfigStore>(msg: &str, uuid: &Uuid, manager: &Manager<C,Regist
...
-                                    attachments_tmp_dir.path().display()
-                                );
+        "attachments will be stored in {}",
+        attachments_tmp_dir.path().display()
+    );
 
-    let messages = manager
-        .clone()
-        .receive_messages()
+    let clone = manager.clone();
```
gferon, line 202: not necessary anymore?

src/config/secret_volatile.rs
```diff
@@ -125,6 +125,39 @@ impl ContactsStore for SecretVolatileConfigStore {
     }
 }
 
+impl MessageStore for SecretVolatileConfigStore {
+    fn save_message(&mut self, _message: libsignal_service::prelude::Content) -> Result<(), Error> {
+        warn!("messages are not saved when using volatile storage.");
+        Ok(())
+    }
+
+    fn messages(&self) -> Result<Vec<libsignal_service::prelude::Content>, Error> {
```
gferon, line 134: I think it would be better if this was `impl Iter<Item = Content>` instead.

gferon (changes requested):
I realized that this is already such a good contribution that I'd rather move forward with it, and add a new storage backend with `sqlite` afterwards.

I have some questions about the data model though. To me, the main use-case is to load threads from a client, not to fetch a single messages. I think it might be a better idea to assign IDs to "threads" from either a group or a single recipient (one call) and then return an iterator over all messages ordered by timestamp for one thread. You could also spare the first lookup entirely by making a composite key the way you did it in [MessageIdentity](https://github.com/whisperfish/presage/pull/67/files#diff-b0ba3e54b6f5ef1fa6aed8075e352a68fda66067fbafa8421042a2347801338aR50). What do you think?

Schmiddiii:
> think it might be a better idea to assign IDs to "threads" from either a group or a single recipient (one call) and then return an iterator over all messages ordered by timestamp for one thread. You could also spare the first lookup entirely by making a composite key the way you did it in [MessageIdentity](https://github.com/whisperfish/presage/pull/67/files#diff-b0ba3e54b6f5ef1fa6aed8075e352a68fda66067fbafa8421042a2347801338aR50). What do you think?

I am not entirely sure what you mean with that. Do you think `messages_by_contact/group` should return `impl Iterator<Item = Content>` instead of `Vec<MessageIdentity>`? Or do you want to combine `messages_by_contact/group` into one `message_by_thread(Thread)` where `Thread` is a enum either Contact or Group?.

Either way, I think returning `impl Iterator<Item = Content>` (or maybe more future-proof for diesel-async `impl Stream<Item = Content>`?) is seems better.

src/config/secret_volatile.rs
Schmiddiii: This is harder then I imagined. Just swapping out `Vec<_>` to `impl Iterator<Item = _>` results in `error[E0562]: `impl Trait` only allowed in function and inherent method return types, not in trait method return`, creating a new type for the returned value works, but one cannot give a concrete value to the type in the implementations as it uses closures for `filter_map`ping. Maybe `Box<dyn Iterator<Item = _>>` is better here?

Schmiddiii:
Just a small FYI, this is now successfully used in a [Flare Draft MR](https://gitlab.com/Schmiddiii/flare/-/merge_requests/12), so it seems to be working fine. And before merging, this should probably be squashed into one commit as it got a little messy commit-wise.

gferon:
Sorry for the radio silence, I see you're still working on this 💪🏼 

> I am not entirely sure what you mean with that

IMO, the main use-case for retrieving messages is for a UI to display what I'll call a "thread". This can be either a 1-1 thread, or a group thread. `sled` has a [https://docs.rs/sled/latest/sled/struct.Tree.html#method.range](range) method on is `Tree` struct, which sounds perfect to me if each thread has its own tree, and where the key is the timestamp of the message. Messages don't have IDs in official Signal clients, and using the timestamp is how reactions are being applied to the correct message, for instance.

At the end of the day we can either have multiple methods like you did in the trait, or something that takes some enum that describes the key. Right now, I'm thinking of something like this:

```rust
enum Thread {
  Contact(Uuid),
  Group(GroupKey),
}

fn messages(thread_key: Thread, from: Option<u64>) -> impl Iter<Msg>;
```

Using the `timestamp` in the retrieving method makes it trivial to paginate if you want to only load some messages for example. It also makes it easy to trim the store for "messages older than X".

I'm sorry this might mean you need to rework some of what you've done here, if that's an option I'm also willing to take it as-is and apply some changes on top.

Schmiddiii:
Now I see what you mean, and it sounds pretty brilliant. There are only a few minor issues/questions:

- A thread can have multiple messages at the same timestep (even though probably very rare). One could maybe check and move the message by 1ms if needed (I would highly doubt that there is a chat room out there that has more than one message per milliseconds, which would lead to problems in that case).
- The pagination sounds really great, but I will need to investigate first if "load the last 10 messages before X timestamp" is possible without loading the entirety of every message before that timestamp (meaning: Does the iterator dynamically load the messages). I would guess they are dynamically loaded, otherwise they probably would not return iterators. I will probably need a closer look at the sled documentation for this.

I will try to investigate the things you mentioned closer the next few days.

gferon:
> A thread can have multiple messages at the same timestamp

While technically true, Signal already takes care of that in the server implementation, and since you receive messages only in the context of a thread it cannot happen. This is an assumption we can make, and we could even return an error on storage in the case where that ever happens (thank you 🦀)

> without loading the entirety of every message before that timestamp

That's exactly the point of the `range()` method I mentioned. `sled` is based on the concept of a LSM Tree with extra optimizations, which perform optimizations on common prefixes (which all timestamps can have - think buckets).

> as for `impl Iterator<Item=Msg>` in a trait, IIRC we can try the approach of returning a struct, which then implement `IntoIterator` or return `Box<dyn Iterator>`

Schmiddiii:
I think everything needed should now be implemented. I have not yet tested it in the CLI, but it works for Flare (and is actually easier to use). Thanks again for the great advice. Could you maybe especially look over the logic of turning `Content+Option<ServiceAddress>` to `Thread` and check if there are some special cases I forgot to consider (already happened to me twice).

src/manager.rs
```diff
@@ -823,3 +823,19 @@ impl<C: ConfigStore> Manager<C, Registered> {
...
+impl<C> Manager<C, Registered>
+where
+    C: ConfigStore + MessageStore,
+{
+    pub async fn receive_messages_store(&self) -> Result<impl Stream<Item = Content> + '_, Error> {
+        let mut store = self.config_store.clone();
+        Ok(self.receive_messages().await?.map(move |c| {
+            if c.metadata.sender.uuid.is_some() {
+                // TODO: Error handling?
+                let _ = store.save_message(c.clone(), None::<ServiceAddress>);
```
gferon, line 836: I think here, you could just log errors, I don't feel like we should interrupt the stream if we can't decode a single message.

src/config/sled.rs
```diff
@@ -460,6 +464,112 @@ impl IdentityKeyStore for SledConfigStore {
...
+    ) -> Result<(), Error> {
+        let receiver = receiver.map(|s| s.into());
+
+        let id = MessageIdentity::try_from(&message)?;
+        let thread = Thread::from_content_receiver(&message, receiver.as_ref())?;
+        let timestamp = id.1;
+
+        log::trace!("Storing a message with id: {:?}", id);
+
+        let tree_messages = self
```
gferon, line 488: Do you have a real need for storing and accessing messages one by one? IMO, we should only store messages once.

src/config/sled.rs
```diff
@@ -460,6 +464,112 @@ impl IdentityKeyStore for SledConfigStore {
...
+        let val = tree.get(key)?;
+        if let Some(val) = val {
+            let proto = ContentProto::decode(&*val)?;
+            Ok(Some(proto.into_content()))
+        } else {
+            Ok(None)
+        }
+    }
+
+    fn messages_by_thread(
```
gferon, line 541: Following the previous comment, I would rename this method `messages` with the same parameters.

examples/cli.rs
```diff
@@ -93,6 +93,18 @@ enum Cmd {
     ListGroups,
     #[clap(about = "List contacts")]
     ListContacts,
+    #[clap(about = "List messages")]
+    ListMessages,
+    #[clap(about = "List messages from contact")]
+    ListMessagesFromContact {
```
gferon, line 99: if you apply the comments from above and only end up having two ways of retrieving messages, you can use a neat trick from `clap_derive` and have a single command, where one argument has an additional attribute `conflicts_with` with the other (just be careful that it should be set to `master-key` and not `master_key`).

examples/cli.rs
```diff
@@ -416,6 +433,83 @@ async fn run<C: ConfigStore>(subcommand: Cmd, config_store: C) -> anyhow::Result
...
+                            Some(Sent {
+                                message: Some(message),
+                                ..
+                            }),
+                        ..
+                    }) => {
+                        println!(
+                            "{}: {}",
+                            msg.metadata.sender.e164_or_uuid(),
+                            message.body.unwrap_or_else(|| "".to_string())
```
gferon, line 453: what about filtering only message that have a body in the `match` statement above? this would prevent the output from being littered with empty "system" messages.

gferon (changes requested):
Exciting! I feel like this is getting close to be ready for merging 🥳 thanks again for your work on this.

PR #68, opened 2022-08-17
Update libsignal-service

Schmiddiii:
This enables the fix in <https://github.com/whisperfish/libsignal-service-rs/pull/119>. I am not completely sure what name to give the device when registering as primary device and I am also unsure what device-capabilities presage has, so I mostly left it at default in combination with the previous device capabilities.

No need to hurry merging.


gferon:
Thanks!

PR #71, opened 2022-09-25
Some cleanups after the last wave of changes

### LINKED ISSUE #69: Start versioning using tags
Since we can't publish our crates' ecosystem on crates.io because we have a few `git` dependencies (mostly `libsignal-client`) we should at least try to version `presage` with tags that are known to be working.

gferon:
Following the last few changes and PRs like #67, I wanted to clean a few things. Here's an exhaustive list:

- [x] Fix some documentation (`receive` > `recipient`)
- [x] Fixes for the update to `libsignal` 0.20.0 (https://github.com/whisperfish/libsignal-service-rs/pull/154)
- [x] Add `delete_message` to `MessageStore`
- [x] Use `Into`, `From`, `TryInto` and `TryFrom` instead of named methods like `ContentProto::from_content`
- [x] **BREAKING CHANGE:** Rename `ConfigStore` to `Store` since it stores more than configuration now
- [x] Remove dead code path in `save_message`
- [x] Bump crate version to 0.4.0 (will be the first tagged version)
- [x] Save outgoing messages in threads

Closes #69


Schmiddiii:
I noticed you deleted the `recipient`-argument from the save-function. But I think this is needed to identify a `Thread` correctly, as messages sent out will not have a recipient-tag attatched. It is obviously still correct as presage only stores messages when receiving, where this case does not happen, but with Flare (and probably all other messengers based on presage), they probably also want to store outgoing messages, where they will need the recipient of the message. Did I miss something that outgoing messages will still be inserted correctly?

gferon:
> Did I miss something that outgoing messages will still be inserted correctly?

You're absolutely right, this is exactly why I left this PR open - I removed the `receiver` parameter because I don't think it's fair to expect users to have to interact with the internal API of the store. I'll add the missing bits.

PR #73, opened 2022-10-05
Implement local encryption

gferon:
Implement encryption of the values stored in the `sled::Db` which is in theory the last missing feature in `presage` 🥳

This breaks again the compatibility of previous sled stores (sorry 😢) since I also took the opportunity to refactor the code and introduce this change with more removed code than added (!)

EDIT: will not merge this before I come up with some sort of schema migration system (which shouldn't be too hard, but will take a bit of extra time)

### DIFF
--- a/src/store/sled.rs
+++ b/src/store/sled.rs
@@ -491,7 +491,7 @@ impl MessageStore for SledStore {
             .db
             .try_read()
             .expect("poisoned mutex")
-            .open_tree(thread_key(&thread))?;
+            .open_tree(thread_key(thread))?;
 
         let timestamp_bytes = message.metadata.timestamp.to_be_bytes();
         let proto: ContentProto = message.into();
--- a/src/store/sled.rs
+++ b/src/store/sled.rs
@@ -23,6 +23,7 @@ use log::{debug, trace, warn};
 use matrix_sdk_store_encryption::StoreCipher;
 use prost::Message;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use sled::Batch;
 
 use super::{ContactsStore, MessageStore, StateStore};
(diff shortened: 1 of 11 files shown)


src/store/sled.rs
```diff
@@ -18,193 +15,196 @@ use libsignal_service::{
...
+    ) -> Result<StoreCipher, Error> {
+        let cipher = if let Some(key) = database.get(SLED_KEY_STORE_CIPHER)? {
+            StoreCipher::import(passphrase, &key)?
+        } else {
+            let cipher = StoreCipher::new()?;
+            #[cfg(not(test))]
+            let export = cipher.export(passphrase);
+            #[cfg(test)]
+            let export = cipher._insecure_export_fast_for_testing(passphrase);
+            database.insert(SLED_KEY_STORE_CIPHER, export?)?;
```
boxdot, line 76: Shouldn't the use store the cipher somewhere? What the point of storing it in sled?

src/store/sled.rs
```diff
@@ -18,193 +15,196 @@ use libsignal_service::{
...
-            .iter()
-            .filter_map(|r| {
-                let (k, _) = r.ok()?;
-                Some(String::from_utf8_lossy(&k).to_string())
-            })
-            .collect();
-        Ok((global_keys, session_keys))
+    /// build a hashed messages thread key
+    fn messages_thread_key(&self, t: &Thread) -> String {
+        format!("{SLED_TREE_THREAD_PREFIX}{t:?}")
```
boxdot, line 143: This should hash at some point, right? Is it for hiding the thread id?

src/store/sled.rs
gferon: This is how it works, akin to how you'd store keys for full disk encryption. What you don't store is the passphrase.

src/store/sled.rs
gferon: Yes, I have a pending commit to deal with some of the keys.

gferon:
> There is a big warning sign in [crates.io](https://crates.io/crates/matrix-sdk-store-encryption) but I think this should be ok.

Just to address this comment, this is exactly used as intended (and is also used in a similar fashion in the `matrix-sdk`). `presage` is not a final application but also a building block.

PR #84, opened 2023-01-04
If no profile_key is provided try to get it from the registered contacts

nanu-c:
the retrieve profile asks for a profile key even when presage has that information already. This patch looks in the registered tree for the profile key.


gferon (approved):
Very nice, thanks a bunch!

PR #85, opened 2023-01-04
Add optional system notifications via

nanu-c:
This commit adds notifications via a generic rust library

### DIFF
--- a/examples/cli.rs
+++ b/examples/cli.rs
@@ -1,4 +1,5 @@
 use core::fmt;
+use std::convert::TryInto;
 use std::{path::PathBuf, time::UNIX_EPOCH};
 
 use anyhow::{anyhow, bail, Context as _};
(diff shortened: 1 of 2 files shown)


nanu-c:
Question, there is no easier way to get my uuid? And i didn't use `manager.get_contact_by_id(uuid)` because it's broken and throws always a `Error: JsonError(Error("invalid type: map, expected a sequence", line: 1, column: 0))`

nanu-c:
I found the `uuid`, but i would recommend to rename it to `my_uuid`. I have update the pr accordingly

examples/cli.rs
```diff
@@ -237,25 +248,75 @@ async fn receive<C: Store + MessageStore>(
...
+                        session_name = group.title;
+                        println!("Group v2: {:?}", session_name);
                         println!("Group change: {:?}", group_changes);
                         println!("Group master key: {:?}", hex::encode(&master_key_bytes));
+                    } else if !self_is_sender {
+                        //get the contact name
+                        match metadata.sender.uuid {
+                            Some(uuid) => {
+                                println!("uuid: {:?}", uuid);
+                                for contact in manager
```
gferon, line 283: I feel like we should find a way to put this inside of `Manager` since IMO this is what `presage` is about. I'm fine merging this as-is for now, though. I'll try to see if there's a way to include it inside.

nanu-c: I agree, but it should be optional, because maybe it's used on a headless server without the need for notifications.

nanu-c:
I have to test if typing messages etc. also trigger a message

gferon:
I ended up rewriting the message receiving part of the CLI to format messages in a much prettier fashion. Thanks for the PR!

PR #87, opened 2023-01-05
update contacts from profile

nanu-c:
In my case I want to have the username as contact name because i don't know who is who in the contacts list.


nanu-c:
This should be in the manager and not in the cli.

gferon:
I'm not sure what the best behavior is here. If you have a contact from your phone, it means that you added somebody to your contacts manually, and I suppose that this should always take precedence over anything. Should we update the actual `Contact` in the local DB or maintain a RAM cache of profiles we fetch on demand?

gferon:
Redone in #211

PR #88, opened 2023-01-06
Add groups storage

nanu-c:
This implements a group store and saves the groups as protobuf messages in the store.


gferon:
Awesome addition, it was on my todo list forever 😄

examples/cli.rs
```diff
@@ -422,7 +423,29 @@ async fn run<C: Store + MessageStore>(subcommand: Cmd, config_store: C) -> anyho
         Cmd::Block => unimplemented!(),
         Cmd::Unblock => unimplemented!(),
         Cmd::UpdateContact => unimplemented!(),
-        Cmd::ListGroups => unimplemented!(),
+        Cmd::ListGroups => {
+            let manager = Manager::load_registered(config_store)?;
+            let mut groups = manager.get_groups()?.peekable();
+            if groups.peek().is_some() {
```
gferon, line 429: I get that you want to print "No groups found" but I would just skip the `peekable` and test entirely. If it prints nothing, then it's empty.

examples/cli.rs
```diff
@@ -422,7 +423,29 @@ async fn run<C: Store + MessageStore>(subcommand: Cmd, config_store: C) -> anyho
...
+                println!("Groups: ");
+                for group in groups {
+                    match group {
+                        Ok(group) => {
+                            let title = String::from_utf8(group.title.clone())?;
```
gferon, line 434:
```suggestion
                            let title = str::from_utf8(&group.title)?;
```

examples/cli.rs
```diff
@@ -422,7 +423,29 @@ async fn run<C: Store + MessageStore>(subcommand: Cmd, config_store: C) -> anyho
...
+                            let key = base64::encode(group.public_key.clone());
```
gferon, line 435:
```suggestion
                            let key = base64::encode(&group.public_key);
```

examples/cli.rs
```diff
@@ -422,7 +423,29 @@ async fn run<C: Store + MessageStore>(subcommand: Cmd, config_store: C) -> anyho
...
+                            println!("{:?} {:?}",key,  title);
```
gferon, line 436:
```suggestion
                            println!("{title} {key}");
```

examples/cli.rs
```diff
@@ -422,7 +423,29 @@ async fn run<C: Store + MessageStore>(subcommand: Cmd, config_store: C) -> anyho
...
+                        }
+                        Err(e) => {
+                            println!("Error: {:?}", e);
```
gferon, line 439: Maybe use the `error!` macro here instead? And avoid using the `Debug` representation of the error as well.

examples/cli.rs
```diff
@@ -476,23 +499,25 @@ async fn run<C: Store + MessageStore>(subcommand: Cmd, config_store: C) -> anyho
             }
         }
         Cmd::GetGroup { group_master_key } => {
+            let( master_key, key ) = group_master_key;
             let manager = Manager::load_registered(config_store)?;
-            let group = manager.get_group_v2(group_master_key).await?;
+            let group = manager.get_group_v2(master_key).await?;
```
gferon, line 504: I'm not super sure why you'd have to do it like this? Is it because you want to use the master key as a string key in the storage backend?

examples/cli.rs
```diff
@@ -476,23 +499,25 @@ async fn run<C: Store + MessageStore>(subcommand: Cmd, config_store: C) -> anyho
...
             println!("{:#?}", DebugGroup(&group));
             for member in &group.members {
                 let profile_key = base64::encode(&member.profile_key.bytes);
                 println!("{member:#?} => profile_key = {profile_key}",);
             }
+            manager.save_group(group, key.as_bytes().to_vec())?;
```
gferon, line 510: I'm pretty sure you can get the bytes representation from the `GroupMasterKey` object directly as well, also without allocating a new `Vec`.

src/manager.rs
```diff
@@ -811,6 +817,26 @@ impl<C: Store> Manager<C, Registered> {
         Ok(group_changes)
     }
 
+    pub fn save_group(&self, group:Group, key:Vec<u8>) -> Result<(), Error> {
```
gferon, line 820:
```suggestion
    pub fn save_group(&self, key: impl AsRef<[u8]>, group: Group) -> Result<(), Error> {
```

src/proto.rs
```diff
@@ -110,3 +111,26 @@ impl TryInto<Content> for ContentProto {
         Content::from_proto(self.content, self.metadata.into()).map_err(Error::from)
     }
 }
+
+#[derive(Clone, PartialEq, ::prost::Message)]
```
gferon, line 115: Maybe I'm confused but I feel like we don't need this wrapper to serialize the group?

gferon (changes requested):

examples/cli.rs
nanu-c: The `proto::group` has no field for the master_key, so how should we store the masterkey? Either by adding a new field or using it as key in the tree.

nanu-c: I am unsure, it was just a shot in the dark. First i looked into the `groups_v2::Group` but this doesn't serialize into json or a protobuf, so I tried it this way.

examples/cli.rs
nanu-c: Actually i would love to use `log::Error` and `log::Info` or `error!` but they don't print anything to my console, so I ended up with `println!` statements. Regarding the `{:?}`, I still have to read some basics about the representations.

examples/cli.rs
nanu-c: Done

examples/cli.rs
nanu-c: i removed the "No groups found"

examples/cli.rs
nanu-c: done

examples/cli.rs
nanu-c: done

examples/cli.rs
nanu-c: the macro `error!` isn't defined and i changed it to an error representation.

src/manager.rs
nanu-c: here we have to rethink, it a little. I set the `public_key: key` which needs an vector. I misuse that field to pass the group id to `self.config_store.save_group(proto_group)?;` We have to either make a struct à la {key, group} or we need to pass the key as second argument.

Schmiddiii:
Some minor comments from my side:

- Should `presage` also have a `request_groups_sync` similar to `request_contacts_sync` (if that is even possible)? 
- And probably the client using `presage` should not have to call `save_group` themselves, but I would expect that to be called inside `presage` (either when syncing groups like previously mentioned or deconstructing `Message` to search for new groups inside `receive_messages`).

Otherwise this `GroupsStore`-trait seems good for usage in Flare (once I update to use the `presage`-store).

nanu-c:
Actually should we add the group sync here or in another pr?

gferon:
> Actually should we add the group sync here or in another pr?

I would take care of this separately, this is already working fine fetching groups when you receive messages.

PR #91, opened 2023-01-10
Fix Manager::contact_by_id

nanu-c:
This wasn't working.


gferon:
@nanu-c the real issue was the double serialization (which was a leftover from the `Store` refactoring I did a while ago). I also added a `get-contact` command to the example CLI to make sure it works as intended.

PR #92, opened 2023-01-11
Use SignalWebSocket

gferon:
This brings a newer version of `libsignal-service-rs` with relevant changes.

Contacts sync will now wait until it is done, like Signal Desktop does.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -1,6 +1,6 @@
 use std::time::UNIX_EPOCH;
 
-use futures::{channel::mpsc, channel::oneshot, future, AsyncReadExt, Stream, StreamExt};
+use futures::{channel::mpsc, channel::oneshot, future, pin_mut, AsyncReadExt, Stream, StreamExt};
 use log::{debug, error, info, trace};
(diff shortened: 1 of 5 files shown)


src/manager.rs
```diff
@@ -536,10 +564,18 @@ impl<C: Store> Manager<C, Registered> {
             .expect("Time went backwards")
             .as_millis() as u64;
 
+        // start waiting for the contacts sync
+        info!("waiting for contacts sync for 3 minutes");
```
Schmiddiii, line 568: Where are the three minutes coming from? I see no such mention in the code.

gferon: True! Just a comment from something I ended up not doing, since I didn't want to introduce a dependency to `tokio`.

Schmiddiii: `presage` already has an indirect dependency on `tokio` in `libsignal-service-hyper`.

gferon: True, I'll do it then as it aligns with the goals of the project.

PR #111, opened 2023-03-04
Fix storing messages

Schmiddiii:
Currently, message timestamps are encrypted when storing. This does not work when retrieving messages before a timestamp (and I think currently does not work when getting/deleting messages in general). This PR fixes this by just removing the encryption.

Before this PR should actually be merged, it should maybe be discussed if removing the encryption is actually a good thing to do, but as far as I see it is the only thing possible except for every time we look for messages of a thread, decrypt all message keys and manually sort and iterate them.

### DIFF
--- a/src/store/sled.rs
+++ b/src/store/sled.rs
@@ -158,16 +158,6 @@ impl SledStore {
         self.db.open_tree(tree).map_err(Error::DbError)
     }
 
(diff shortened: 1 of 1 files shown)


gferon (approved):
This was quite the oversight, I did this mechanically when implementing store encryption and didn't use it in my clients. I'll add some tests soon. Thanks for fixing it!

PR #112, opened 2023-03-04
Fix clearing does not unlink

Schmiddiii:
Don't forget to flush!

Also, I am not sure if `STORE_CIPHER` should be removed or not, as if it is removed, contacts, groups and messages may not be read anymore.

### DIFF
--- a/src/store/sled.rs
+++ b/src/store/sled.rs
@@ -195,6 +195,7 @@ impl SledStore {
             |c| c.encrypt_value(&value).map_err(Error::from),
         )?;
         let _ = self.tree(tree)?.insert(key, value)?;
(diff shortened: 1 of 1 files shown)


Schmiddiii:
Thinking about it, maybe we should introduce a `flush`-method to the stores, as theoretically one may loose some data when closing an application. Practically I don't think that happens that often as `sled` flushes itself twice a second (I think).

gferon:
This is something that I've seen when dealing with `sled`. Some libraries call `.flush()` on every single write call they do to `sled` (like the Matrix SDK). One alternative would be to add an exit handler (to exit gracefully).

PR #113, opened 2023-03-04
Fix contacts syncing after linking

Schmiddiii:
This previously returned `MessagePipeNotStarted` as the message pipe was not yet started when sending the request for contact syncing. This behavior was introduced in #106.


gferon:
Thanks for catching and fixing this issue!

PR #115, opened 2023-03-05
Clear messages

Schmiddiii:
This (ab)uses the fact that all thread-trees start with `thread`. I don't know if that should be seen as an interpretation detail of `messages_thread_tree_name` and should therefore be clasified as a workaround. Another option may be to instead query possible `Thread` first and then calculate the thread name, but this would tightly couple the three Store-traits and one would be careful in which order to delete the data.


gferon:
Looks perfectly fine to me, it's not abusing the implementation detail IMO because it's encapsulated inside of the `sled` implementation of the `MessageStore`.

gferon (approved):

PR #119, opened 2023-03-08
Accept ranges when fetching messages from store

### LINKED ISSUE #116: Add basic store tests
Just to make sure we never break fetching messages from store, and can retain the expectations from the public API.

gferon:
and add store tests

Closes #116 

cc: @Schmiddiii

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Changed (and fixed) the behaviour of the iterator returned by `SledStore::messages`:
+- Changed (and fixed) the behaviour of the iterator returned by `SledStore::messages` (#119)
   * The iterator yields elements in chronological order (used to be reversed)
   * The iterator now implements `DoubleEndedIterator` which means you it can be reversed or consumed from the end
   * The method now accepts the full range syntax, like `0..=1678295210` or `..` for all messages
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # Changelog
 
-0.5.0-dev
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Changed
 
 - Changed (and fixed) the behaviour of the iterator returned by `SledStore::messages`:
-  - The iterator yields elements in chronological order (used to be reversed)
-  - The iterator now implements `DoubleEndedIterator` which means you it can be reversed or consumed from the end
-  - The method now accepts the full range syntax, like `0..=1678295210` or `..` for all messages
+  * The iterator yields elements in chronological order (used to be reversed)
+  * The iterator now implements `DoubleEndedIterator` which means you it can be reversed or consumed from the end
+  * The method now accepts the full range syntax, like `0..=1678295210` or `..` for all messages
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+0.5.0-dev
+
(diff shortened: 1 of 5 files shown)


src/store/sled.rs
```diff
@@ -805,33 +804,43 @@ impl MessageStore for SledStore {
...
+        range: impl RangeBounds<u64>,
+    ) -> Result<Self::MessagesIter, Error> {
         let tree_thread = self.tree(self.messages_thread_tree_name(thread))?;
         debug!("{} messages in this tree", tree_thread.len());
-        let iter = if let Some(from) = from {
-            tree_thread.range(from.to_be_bytes()..)
-        } else {
-            tree_thread.range::<&[u8], std::ops::RangeFull>(..)
+
+        let iter = match (range.start_bound(), range.end_bound()) {
```
Schmiddiii, line 815: That looks pretty ugly. Would be useful if [`Bound::map`](https://doc.rust-lang.org/nightly/core/ops/enum.Bound.html#method.map) was stable.

src/store/sled.rs
```diff
@@ -805,33 +804,43 @@ impl MessageStore for SledStore {
...
+            (Bound::Included(start), Bound::Excluded(end)) => {
+                tree_thread.range(start.to_be_bytes()..end.to_be_bytes())
+            }
+            (Bound::Included(start), Bound::Included(end)) => {
+                tree_thread.range(start.to_be_bytes()..=end.to_be_bytes())
+            }
+            (Bound::Unbounded, Bound::Included(end)) => tree_thread.range(..=end.to_be_bytes()),
+            (Bound::Unbounded, Bound::Excluded(end)) => tree_thread.range(..end.to_be_bytes()),
+            (Bound::Unbounded, Bound::Unbounded) => tree_thread.range::<[u8; 8], RangeFull>(..),
+            (Bound::Excluded(_), _) => unreachable!("range that excludes the initial value"),
```
Schmiddiii, line 826: Why should it not be possible to exclude the first value? E.g. when loading messages later than a message one already has but withou that message, excluding the starting-value would be useful.

CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
```
Schmiddiii, line 1: Keeping a changelog would indeed be nice. How about a more common format like shown e.g. [here](https://keepachangelog.com/en/1.1.0/) (which I also use for Flare).

src/store/sled.rs
gferon: It's not about what's being useful or not, it's about what one can express with the `a..b` syntax in Rust. AFAIK, you cannot express this.

src/store/sled.rs
gferon: `Bound::map` wouldn't help much here, since the `.range` function takes a `RangeBounds` as well.

src/store/sled.rs
Schmiddiii: You can use that `RangeBound` is implemented for `(Bound, Bound)` (https://docs.rs/rustc-std-workspace-std/1.0.1/std/ops/trait.RangeBounds.html#foreign-impls). I also do that for Flare.

src/store/sled.rs
Schmiddiii: You could just use `(range.start_bound().map(|b| b.to_be_bytes()), range.end_bound().map(|b| b.to_be_bytes())` as bounds to the sled iterator if there would be a `map` I think. But, as mentioned, nothing you have to fix.

src/store/sled.rs
gferon: This is nice, thanks (but we still have to do things ourselves here until the method is stabilized).

Schmiddiii: What I have mentioned here is already stable. But as mentioned, nothing really necessary, just a minor comment.

CHANGELOG.md
Schmiddiii: You forgot to link the `[Unreleased]`-section to the GitHub commits. But nothing really important.

src/store/sled.rs
gferon: Note: because `Bound::map` is not there yet (and there doesn't seem to be any news on that front in https://github.com/rust-lang/rust/issues/86026) you can see that `sled` does something similar https://docs.rs/sled/latest/src/sled/tree.rs.html#1207

PR #130, opened 2023-03-10
Attachment Store for caching attachments

Schmiddiii:
This enables caching of attachments. Some notes:

- This assumes an `AttachmentIdentifier` of an `AttachmentPointer` uniquely references an attachment which may not change.
- We may also want to cache attachments when uploading.
- We may also want a method checking if there is a local version of the attachment.
- We probably want something to delete old attachments (similar to LRU-cache). Not yet sure how this can be implemented, but maybe one could add the timestamp of the save to the stored attachment and then iterate over all attachments (`O(n)`) and delete the old ones.


Schmiddiii:
I feel like this pR is too outdated with the current main branch. Even though this is still something I want some day, I think it is better to close this PR and redo it on a later day.

PR #131, opened 2023-03-10
Delete messages

Schmiddiii:
Things to think about:

- Do not really delete from store, but mark as deleted and detele information. This would probably require a pretty different store of the messages though.
- Clients still have to react to delete-messages to dynamically delete showed messages. 
- The giant match-clauses for the different message-types are getting pretty big. Maybe somehow refactor.


gferon:
Thanks! I took a slightly different approach in #147 and replace deleted messages by `NullMessage`.

PR #132, opened 2023-03-11
Store for receipts

Schmiddiii:
~~Currently not too happy with the actual implementation of the store and also not sure about its exact interface. Will open it either way as a WIP.~~

Updated the trait and implementation to be a little bit better.


gferon:
I gave this a quick look, and I think the API should mirror the one from `Store::message`. I also think we should introduce a struct for `Receipt` instead of implementing the byte representation ourselves. I can work on this to push it to the end if that's ok.

gferon:
(sorry, greasy fingers, reopened.)

Schmiddiii:
I feel like this pR is too outdated with the current main branch. Even though this is still something I want some day, I think it is better to close this PR and redo it on a later day.

PR #134, opened 2023-03-18
Cache for profile

Schmiddiii:
~~Dependent on upstream PR <https://github.com/whisperfish/libsignal-service-rs/pull/207>. Therefore, my fork is currently uses but should be changed to upstream as soon as merged there.~~

This is very similar to <https://github.com/whisperfish/presage/pull/130> and provides pretty much the same functionality for profiles.

Maybe after both are merged, a function `clear_cache` would be useful for the stores.


gferon:
There's no obvious way to force the refresh of profile is we do like this, but it's already better than nothing as it avoids rate-limiting!

PR #142, opened 2023-03-25
Fix duplicate message to group

Schmiddiii:
Currently, when sendinging a message to a group, this message will be sent both as a data message to linked devices and as a sync message. As discussed in the whisperfish-chat, the recipients-list should not contain the own UUID. As also discussed, this breaks sending to groups with only the sender and no other persons, this has to be fixed upstream.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -893,7 +893,12 @@ impl<C: Store> Manager<C, Registered> {
             return Err(Error::UnknownGroup);
         };
 
-        let recipients: Vec<_> = group.members.into_iter().map(|m| m.uuid.into()).collect();
+        let recipients: Vec<_> = group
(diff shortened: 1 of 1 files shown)


gferon:
Thanks! Would you rather merge the other fix upstream first or this one? (I have no preference).

Schmiddiii:
As the upstream fix would contain code that is not really equivalent to the Signal Android code, I did not yet make the PR. Also, Ruben mentioned he works on sealed sending (the new code path), which should probably also fix the upstream. But if you want, I can create the PR.

gferon:
I'll merge this for now and think of how to deal with the remaining issue you mentioned.

gferon (approved):

PR #145, opened 2023-03-31
Fix sending messages to yourself in a group

boxdot:


gferon:
Thanks, great minds think alike, this was already submitted by @Schmiddiii in #142

PR #146, opened 2023-04-01
Sqlite store

Schmiddiii:
A very early draft of a Sqlite store. Things that still need to be done:

- [ ] Encryption of the store.
- [x] Not yet sure if 1-to-1 messages work. Note-to-self seems to work though.
- [ ] Group store (currently, sending or receiving from groups is not possible).
- [ ] Messages store.
- [ ] Using the `src/model.rs`-structs in the stores instead of protobufs (will also require all other store implementations to change).
- [ ] Think about alternatives to returning iterators for messages and others as this is not well supported by diesel.
- [ ] Add metadata to the new crate.
- [ ] Remove `libsignal_service` dependency of store by exporting more in presage. Maybe a `presage::prelude::store`?
- [ ] `down.sql`
- [ ] Better error messages. How can we include storage-errors in general presage errors?
- [ ] Is everything in `src/model.rs` required?


Schmiddiii:
I feel like this PR is too outdated with the current main branch. Even though this is still something I want some day, I think it is better to close this PR and redo it on a later day.

PR #149, opened 2023-04-03
Improved model and stores

Schmiddiii:
An improved model with changes to the stores based on Whisperfish. Currently done:

- Recipients
- Group (only V2)
- Conversation

General things to talk about:

- Do we want for some things a "light core and heavy wrapper", e.g. in
  regards to avatars, these will be pretty expensive to load and (maybe)
  only rarely needed. Therefore, a "light core" could hold the base
  information and the heavy wrapper all information required (e.g.
  avatar, group members (which also require multiple DB accesses to load),
  message receipts/reactions (similar))

This obviously does not compile yet.

Full changelog (and TODOs):

- Added `Timestamp = u64` (and similar):
    - Maybe make to `NaiveDateTime`?

- General `Store`:
    - Defined `clear_data` based on other `clear_*`.
    - Defined `clear` based on `clear_registration` and `clear_data`.

- `Contact`:
    - Renamed to `Recipient` in accordance to WF.
    - Made `Recipient::uuid` from `Option<String>` to `Uuid`.
    - Removed `Recipient::email`.
    - Added `Recipient::archived`.
    - Made `Recipient::e164` from `Option<String>` to `Recipient::phone_number` `Option<PhoneNumber>`.
    - Added `Recipient::color` as `Option<String>` (TODO: Maybe other type?).
    - TODO: `Recipient::verified` similar to `Contact::verified`?
    - `Recipient::profile_key` from `Option<Vec<u8>>` to `[u8; 32]` (TODO: Maybe other type?).
    - TODO: `Recipient::profile_key_credentials` from WF?
    - Added `Recipient::profile` instead of `profile_*_name` and `about_*`.
    - TODO: `signal_profile_avatar`, `profile_sharing`, `last_profile_fetch` from WF?
    - TODO: `unidentified_access_mode`, `storage_service_id`, `storage_proto` from WF?
    - Removed `last_gv1_migrate_reminder`
    - TODO: `capabilities`, `last_session_reset`?
    - TODO: Avatar data?
    - Add `expire_timer` (TODO: Better type? Maybe `Duration`?)

- `ContactsStore`:
    - Renamed to `RecipientsStore` along with all functions.
    - Implementation of `store_recipients` based on `store_recipient`. (TODO: Implementation)
    - `contacts` returns `Vec` instead of `Iterator`.
    - Renamed `by_id` to `by_uuid`

- `Group`:
    - TODO: Maybe split up into inner and augmented group?
    - Made `id` to `[u8; 16]` (TODO: Correct?)
    - Made `master_key` to `[u8; 32]` (TODO: Correct?)
    - Made `invite_link_password` to `Option<String>` (TODO: Correct?)
    - Made `access_*` to `access` of type `GroupAccessControl` inspired by `proto::AccessControl`.
    - Add `expire_timer` (TODO: Better type? Maybe `Duration`?)
    - Add `announcements_only`.
    - Add `members`, `pending_members`, `banned_members` with corresponding types.
    - TODO: `RequestingGroupMember::presentation`?
    - TODO: Avatar?

- `GroupsStore`:
    - `groups` returns `Vec` instead of `Iterator`.
    - `save_group` does not take the master key anymore as it is already included in the struct and `group_id` should be used instead

- `Conversation`:
    - Defined as enum based on group or recipient.
    - TODO: Any other data worth adding?
    - Added conversion to `ConversationId` for usage in `MessagesStore`.

- `MessageStore`:
    - Take `MessageId` instead of two parameters.
    - `save_message` only takes message not including `Conversation` as this may be infered.
    - Predefined methods for marking a message as deleted, adding a reaction or receipt.

- `ProfilesStore`:
    - Not needed anymore as already included in the `ContactsStore`.
    - Maybe a `by_profile_key` required?

- TODO: `AttachmentsStore`, `StickersStore` (similar to WF)
- TODO: `impl Into` for the model structs into the protobuf structs and back.

### DIFF
--- a/presage/src/lib.rs
+++ b/presage/src/lib.rs
@@ -1,6 +1,7 @@
 mod cache;
 mod errors;
 mod manager;
+mod model;
 mod serde;
 mod store;
 
@@ -11,6 +12,7 @@ pub use store::{Store, StoreError, Thread};
 #[deprecated(note = "Please help use improve the prelude module instead")]
 pub use libsignal_service;
 
+pub use model::*;
 pub mod prelude {
     pub use libsignal_service::{
         configuration::SignalServers,
--- a/presage/src/model.rs
+++ b/presage/src/model.rs
@@ -1,0 +1,633 @@
+use std::{collections::HashSet, fmt::Display};
+
+use crate::{prelude::proto, Error};
+use libsignal_service::{
+    content::{ContentBody, DataMessageFlags},
+    prelude::{phonenumber::PhoneNumber, Content, Uuid},
+    proto::{
+        access_control::AccessRequired,
+        data_message::{Contact, Sticker},
+        member::Role,
+        receipt_message::Type,
+        sync_message::Sent,
+        AttachmentPointer, DataMessage, GroupContextV2, Preview, SyncMessage,
+    },
+    Profile,
+};
+
+use libsignal_service::groups_v2 as lssg;
+
+pub type Timestamp = u64;
+pub type ProfileKey = [u8; 32];
+pub type GroupId = [u8; 16];
+pub type GroupRevision = u32;
+pub type MessageId = (Timestamp, ConversationId);
+
+/// A user of Signal.
+#[derive(Debug, Clone)]
+pub struct Recipient {
+    pub uuid: Uuid,
+    pub phone_number: Option<PhoneNumber>,
+    /// The name you gave this user.
+    pub username: Option<String>,
+
+    pub blocked: bool,
+    pub archived: bool,
+
+    pub color: Option<String>,
+
+    pub profile_key: ProfileKey,
+    // pub profile_key_credential: Option<Vec<u8>>,
+    /// The profile of this user including the information he published himself.
+    pub profile: Option<Profile>,
+    // pub signal_profile_avatar: Option<String>,
+    // pub profile_sharing: bool,
+
+    // pub last_profile_fetch: Option<NaiveDateTime>,
+    // pub unidentified_access_mode: bool,
+    // pub storage_service_id: Option<Vec<u8>>,
+    // pub storage_proto: Option<Vec<u8>>,
+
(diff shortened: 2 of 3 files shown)


Schmiddiii:
We plan to define suitable `impl Into` for the models. So what you will probably need to do is call `.into()` on any return type of the store and also some of the manager.

Also, I would not say presage is usable in production yet. While it mostly seems to work, there are still some weird issues sometimes and as this is still pre-1.0 we can make breaking changes as much as we want (not that we plan to do many, but there are still some (including this one) that we plan to do).

gferon:
> Sorry if this is a dumb question, but for those of us using presage in prod, will it auto-migrate the existing storage, or provide a way to do so easily?

First of all, let me reassure you that there is no dumb question 😄 We'll try to have automatic migrations for the storage in place as best as we can. I do not expect this change to be breaking in terms of storage, but you will have to change your client code (a little bit).

k-bx:
All right, thanks for the info. My production usage is extremely simple, basically reading all messages and storing them in postgres, but the system relying on this is pretty important so wanted to check I don't break it badly by accident.

gferon:
> All right, thanks for the info. My production usage is extremely simple, basically reading all messages and storing them in postgres, but the system relying on this is pretty important so wanted to check I don't break it badly by accident.

Alright, feel free to ping me on Matrix so we can maybe discuss a bit more and make sure we don't break anything that you rely on by mistake.

Schmiddiii:
I have now decided to rebase it against master while squashing everything down and also remove the changes that were done to the manager as it would be a not-so-nice time keeping those changes up-to-date with the main branch. Instead, I would recommend just working on the store for now and once we decided it is good, then work on porting the manager. This will also mean that the CI will keep failing for quite some time.

Schmiddiii:
I feel like this PR is too outdated with the current main branch. Even though this is still something I want some day, I think it is better to close this PR and redo it on a later day.

Issue #128, opened 2023-03-09
Threads iterator


gferon:
Right now, one can derive the `Thread` from a message they receive, and decide to fetch all messages from this thread. This works, but doesn't provide any nice way to list all threads.

The idea is to provide what you need to build the main screen of a messaging app/client (like `Signal-Android`).

Idea: it should give you a thread, the picture, the name, the last message, the time of the last message and a handle to the iterator for all its messages

nanu-c:
Mhm in Axolotl we do not maintain a different database rather than sled in presage. Actually what we need is a fast way to assemble a dataset for each thread with:

* the thread name aka contact or group name
* Last message with timestamp
* unread message counter
* Muted boolean
* Self destruction message setting (not for the first view, but as common information through all threads that isn't present in the `Contact`

In the current state i need to decrypt all groups, iterate through all threads to get the last message, sort the threads by the timestamp of the message, maintain a fork of presage for the muted boolean and the unread message counter. As i test this on my main signal account with now about 50 groups and 200 contacts, it takes 3-7 seconds for the initial loading of the "dataset" to show the first view.

PR #157, opened 2023-04-14
Get title for thread

nanu-c:
This adds a function for the manager to get the title for a thread.


gferon:
Thanks for the PR, I think I would rather wait for when we finish working on #149 and/or actually introduce a real `Thread` struct that contains an interator of messages (#128).

gferon:
Since I couldn't make any progress on #149, I don't see why I should block this any further. Thanks for the patch!

PR #161, opened 2023-04-19
Store decrypted groups

### LINKED ISSUE #122: Save decrypted groups
Since the `sled` store can optionally be encrypted, it's a good idea to store the decrypted version of groups to save on resources when using them (they're currently decrypted whenever you fetch them). This is especially an issue on mobile devices.

cc @nanu-c

gferon:
* This meant a much higher CPU usage than expected when loading groups often as we would decrypt groups whenever they're accessed.
* Since the store can be optionally encrypted with a passphrase, this would also means the data was encrypted twice.

cc: @nanu-c 

Closes #122

### DIFF
--- a/presage-store-sled/src/lib.rs
+++ b/presage-store-sled/src/lib.rs
@@ -86,13 +86,13 @@ pub enum SchemaVersion {
     #[default]
     V0 = 0,
     V1 = 1,
-    /// the current version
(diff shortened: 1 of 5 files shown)


gferon:
@nanu-c a few months after you asked for it, but better late than never!

PR #172, opened 2023-05-10
manager: continue when skipping internal msgs

ccstolley:
I noticed that sometimes `receive_messages()` will terminate instead of blocking for new messages. When internal messages aren't desired, the closure returns `None`. I believe this causes `futures::stream::unfold()` to return `Poll::Ready(None)` which will close the stream. I don't think that's what we want.

I suggest we just continue looping when skipping internal messages instead of returning None.


gferon:
Thanks a lot for this!

Issue #171, opened 2023-05-08
Support partial contacts update requests


julianfairfax:
Add support for requesting a partial contacts update from the main Signal device, to be used in clients such as Flare, for contact updates.

gferon:
Just to make sure I understand correctly, would that be similar to https://github.com/whisperfish/presage/issues/98?

julianfairfax:
> Just to make sure I understand correctly, would that be similar to #98?

I suppose yes. I was referring to how Signal-Desktop syncs the contacts on each launch though.

PR #173, opened 2023-05-19
Set the expire timer

### LINKED ISSUE #93: Timed Messages
Currently, sending a message in a channel with timed messages turns off timed messages (see [here](https://gitlab.com/Schmiddiii/flare/-/issues/40)). Instead, the `DataMessage.expire_timer` has to be set to the current length of timed messages (as discussed with @nanu-c on Matrix).

Schmiddiii:
This sets the expire timer of a message such that the current expiration timer of a conversation stays the same (except in cases where it was explicitly set to another value).

It may be a little counterproductive when sending a message to (e.g.) a contact to query the timer for the thread which unwraps it back to querying the contact, but this method may potentially also be used to display the expiration timer.

Note that due to #171, I would guess that the expiration timer always stays the same for contacts directly after linking, e.g. changing it on an official client and then sending a message from presage, presage will set the expiration timer back to before the change.

This fixes #93.

P.S.: The panic when the group bytes are not 32 bytes in length is maybe not that good. Is there a reason for the parameter not being `[u8; 32]`?

### DIFF
--- a/presage/Cargo.toml
+++ b/presage/Cargo.toml
@@ -6,8 +6,8 @@ authors = ["Gabriel Féron <g@leirbag.net>"]
 edition = "2021"
 
 [dependencies]
-libsignal-service = { git = "https://github.com/whisperfish/libsignal-service-rs", rev = "c2f70ef" }
-libsignal-service-hyper = { git = "https://github.com/whisperfish/libsignal-service-rs", rev = "c2f70ef" }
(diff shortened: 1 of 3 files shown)


gferon:
Nice! Thanks for the patch.

> The panic when the group bytes are not 32 bytes in length is maybe not that good. Is there a reason for the parameter not being [u8; 32]?

It's like that for backwards compatibility reasons, because `Manager::send_message_to_group` was introduced _before_ the `GroupMasterKeyBytes` type got introduced in `libsignal` 😢

PR #174, opened 2023-05-19
Introduce capability to import an existing registration

Blackoverflow:
This should add the capability to import an existing registration.

I don't have deeper understanding of presage and only build what looked right.

Please comment if this is in the right direction and point out issues.

The examples do not compile, because SledStore is no module of the crate `presage` (anymore?).

### DIFF
--- a/presage/src/lib.rs
+++ b/presage/src/lib.rs
@@ -5,7 +5,7 @@ mod serde;
 mod store;
 
 pub use errors::Error;
-pub use manager::{Confirmation, Linking, Manager, Registered, Registration, RegistrationOptions};
+pub use manager::{Confirmation, Linking, Manager, Registered, Registration, RegistrationOptions, RegistrationInfo};
 pub use store::{Store, StoreError, Thread};
 
 #[deprecated(note = "Please help use improve the prelude module instead")]
--- a/presage/src/manager.rs
+++ b/presage/src/manager.rs
@@ -91,6 +91,23 @@ pub struct Confirmation {
     password: String,
 }
 
+pub struct RegistrationInfo {
+    phone_number: PhoneNumber,
+    service_id_aci: Uuid,
+    service_id_pni: Uuid,
+    password: String,
+    signaling_key: [u8; 52],
+    registration_id: u32,
+    aci_public_key: PublicKey,
+    aci_private_key: PrivateKey,
+    profile_key: ProfileKey,
+    pni_registration_id: Option<u32>,
+    pni_public_key: Option<PublicKey>,
+    pni_private_key: Option<PrivateKey>,
+    device_name: Option<String>,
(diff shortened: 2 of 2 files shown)


Schmiddiii:
Just a few ideas:

I think it would be better to import that directly into the store instead of into the manager. The flow would therefore be:

- Create an empty `SledStore`.
- Import the registration into the `SledStore`.
- Create the `Manager` from the new store (use `from_registered`).
- Maybe some `Manager`-method to repopulate the contacts and other things in the new store.

I think it would also be even better to instead of just importing the registration to create a general "import from some store into another new store", and then not only import the registration but also contacts, messages, .... But that would certainly be more work.

Blackoverflow:
Thank you for your feedback!

I heavily borrowed from `link_secondary_device` and `confirm_verification_code` which are the other two functions which create registrations.

I think the registration of pre keys and setting of account capabilities are also vital parts of the import. Are there other things to do in order to get a working registration? 

I understand your idea is to replace the block
```rust
let state = Registered::from_import(signal_servers, registration_info);

let rng = StdRng::from_entropy();

let mut manager = Manager {
    rng,
    config_store,
    state,
};

manager.config_store.save_state(&manager.state)?;
```
with
```rust
let state = Registered::from_import(signal_servers, registration_info);

config_store.save_state(&state)?;

let mut manager = Manager::load_registered(config_store)?;
```
correct?

Regarding the import of contacts and messages:
I had to start somewhere and prefer to work in small iterations.
When this MR is done, I continue with messages and at last contacts (because contacts can also be rediscovered).
Importing groups would also be useful, am I right?

Schmiddiii:
What I would suggest to do is something like:

```rust
let old_store = AStoreBasedOnYourCurrentData::new(some_old_path);
let new_store = SledStore::new(some_new_path, optional_passphrase);
migrate_stores(old_store, new_store);
let mut manager = Manager::load_registered(new_store)?;
```

Here `AStoreBasedOnYourCurrentData` is an implementation of `presage::Store` (possibly read-only if you don't plan to use that anymore) and `migrate_stores` takes all the data from the `old_store` (as a start, only the registration data, but can be expanded in the future to include everything else) and copies it into `new_store` (where the parameters are just general `presage::Store`, which can then be used to construct the manager.

While certainly more work then what you currently are doing, this will make general migration between stores possible, e.g. also in the future from `SledStore` to `SqliteStore` or anything else.

> I think the registration of pre keys and setting of account capabilities are also vital parts of the import. Are there other things to do in order to get a working registration?

I am not that familiar with the Signal protocol, I would think that you will want to migrate:

- Registration-struct (state)
- (Signed) Pre-Keys (including offset)
- (Sub-device) Sessions
- Identity key pair
- Sender Keys

PR #177, opened 2023-06-06
fix: short-circuit sending zero attachments

boxdot:
Sending attachments creates a new messages sender, which has several side effects. In particular, it might connect to a socket. Now, we avoid it, in case there are no attachments to send.


gferon:
This might actually be one of the main issues we have with sending messages! Thanks for figuring it out.

PR #183, opened 2023-08-08
Fix registration

gferon:
After merging https://github.com/whisperfish/libsignal-service-rs/pull/238


MarcusGrass:
I made a small PR into your branch https://github.com/whisperfish/presage/pull/184

PR #185, opened 2023-08-09
Registration fixes completion

MarcusGrass:
Closed https://github.com/whisperfish/presage/pull/184 in favor of this, should just push that directly to main.


nanu-c:
This works also for axolotl.

gferon:
Thanks a lot @MarcusGrass for the fixes! 🥳

PR #188, opened 2023-09-14
Upgrade to libsignal-protocol 0.32

boxdot:


boxdot:
Based on https://github.com/whisperfish/libsignal-service-rs/pull/245. Can be merged when this PR lands.

Schmiddiii:
Am I missing something or does this not update the libsignal-commit to the `update-libsignal-0.32`-branch?

Schmiddiii:
I added that patch but it changed nothing.

gferon:
Thanks for the patches @boxdot, let's 🚀 it and squash bugs at they arise.

PR #189, opened 2023-09-20
Add ThreadMetadata store

nanu-c:
This adds a ThreadMetadata store and adds automatic contact creation for messages from unknown senders


presage/src/manager.rs
```diff
@@ -953,6 +959,65 @@ impl<C: Store> Manager<C, Registered> {
                                         log::trace!("{group:?}");
                                     }
                                 }
+                                let uuid = content.metadata.sender.uuid.clone();
+                                match state.config_store.contact_by_id(uuid) {
```
gferon, line 963: I feel like the feature in itself is a good addition, but I would maybe do this with a function like `upsert_contact` similar to `upsert_group`.

presage/src/manager.rs
```diff
@@ -953,6 +959,65 @@ impl<C: Store> Manager<C, Registered> {
...
+                                                    destination_uuid: None,
+                                                    null_message: None,
+                                                },
+                                                blocked: false,
+                                                expire_timer: 0,
+                                                inbox_position: 0,
+                                                archived: false,
+                                                avatar: None,
+                                                profile_key: profile_key.to_vec(),
+                                            };
```
gferon, line 989: I guess most of the fields here are default values to you could try to only set `uuid` and add at the end

```
..Default::default()
```

presage/src/manager.rs
```diff
@@ -953,6 +959,65 @@ impl<C: Store> Manager<C, Registered> {
...
+                                    Err(e) => {
+                                        log::error!("Error getting contact: {}", e);
+                                    }
+                                }
+
+                                let thread = Thread::try_from(&content).unwrap();
+
+                                let store = &mut state.config_store;
+
+                                match store.thread_metadata(&thread) {
```
gferon, line 1010: weird that you can't do `state.store.thread_metadata`?

presage/src/manager.rs
```diff
@@ -953,6 +959,65 @@ impl<C: Store> Manager<C, Registered> {
...
+                                    Ok(metadata) => {
```
gferon, line 1011:
```suggestion
Ok(None) => {
```

gferon (commented):
Nice addition! I have a bunch of comments. Would you mind if I play around with this PR and potentially integrate the metadata inside of something that would be called `Thread` (we would rename the current one into `ThreadKey`) and contains a `MessagesIter` as well?

nanu-c:
For me it's perfectly fine if you take this as a base and rename it to Thread. Therefore should i solve the comments?

gferon:
I'll take care of the comments, you don't have to worry about it. Thanks for the initial push!

gferon:
Just wanted to drop a comment to say I've not forgotten this PR and started to work on it. It involves some clean up of the profiles and contacts structs to do things properly. This also means we'll be able to add a bit more data inside the store, such as avatars for groups and contacts 🥳

nanu-c:
Is there something i can do to help you?

gferon:
> Is there something i can do to help you?

I'll pick it back up next week :-) I was attending Eurorust with @rubdos in the past few days

presage/src/manager.rs
nanu-c: the trait `std::default::Default` is not implemented for `libsignal_service::models::Contact`.

presage/src/manager.rs
nanu-c: You are right i can use the `state.config_store.thread_metadata`.

nanu-c:
> I'll pick it back up next week :-) I was attending Eurorust with @rubdos in the past few days
That sounds nice :) I keep experimenting with this branch to find out if something is missing. And for example I need the `expire_timer`, too.

darkdragon-001:
@gferon any updates from your side?

PR #190, opened 2023-09-23
Replace matrix-sdk-store-encryption by our own store cipher

boxdot:
Even though, the implementation of `presage_store_cipher::StoreCipher` is almost verbatim along the lines in `matrix-sdk-store-encryption`, our custom implementation allows us to decouple from dependencies of matrix and their release process.

Also fix `encryption` feature flags: now it is possible to use presage-store-sled without encryption.

### DIFF
--- a/presage-store-sled/src/lib.rs
+++ b/presage-store-sled/src/lib.rs
@@ -121,6 +121,11 @@ impl SledStore {
             .map(|p| Self::get_or_create_store_cipher(&database, p.as_ref()))
             .transpose()?;
 
+        #[cfg(not(feature = "encryption"))]
+        if passphrase.is_some() {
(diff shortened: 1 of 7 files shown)


boxdot:
I think it makes sense to add another build job which build `presage-store-sled` without default flags.

gferon:
This is actually a good idea to use a trait. Let's file an issue?

presage-store-sled/src/lib.rs
```diff
@@ -109,6 +109,7 @@ impl SchemaVersion {
 }
 
 impl SledStore {
+    #[allow(unused_variables)]
```
gferon, line 112: I guess either we can switch to providing an `impl` or `StoreCipher` like you suggested, or at least return an error if this `passphrase.is_some()` and the `encryption` feature is not enabled?

PR #192, opened 2023-09-29
Fix registration without force

Schmiddiii:
In the presage-store-sled implementation, `load_state` always returns `Ok(_)` (except for errors). Therefore, the registration condition that determines if the store was already registered, `!force && config_store.load_state.is_ok()` is always `true` when `force == false`. Instead, one should use `config_store.is_registered()` to figure out if the store is already registered.


gferon:
Nice find, thanks!

PR #193, opened 2023-10-05
Link secondary devices as primary device

Schmiddiii:
Currently does not work due to an issue on the secondary-side:

```
ProvisioningError(InvalidData { reason: "missing client UUID" })
```

The code was mostly copied from what Whisperfish is doing.

Depends on https://github.com/whisperfish/libsignal-service-rs/pull/250

### DIFF
--- a/presage-cli/src/main.rs
+++ b/presage-cli/src/main.rs
@@ -514,7 +514,9 @@ async fn run<S: Store>(subcommand: Cmd, config_store: S) -> anyhow::Result<()> {
                 async move {
                     match provisioning_link_rx.await {
                         Ok(url) => {
-                            qr2term::print_qr(url.to_string()).expect("failed to render qrcode")
+                            println!("Please scan in the QR code:");
+                            qr2term::print_qr(url.to_string()).expect("failed to render qrcode");
+                            println!("Alternatively, use the URL: {}", url);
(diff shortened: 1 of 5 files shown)


Schmiddiii:
The above mentioned issue seems to be due to PNI fields not being set from the primary-device side [here](https://github.com/whisperfish/libsignal-service-rs/blob/main/libsignal-service/src/account_manager.rs#L329). So I guess this MR is blocked by https://github.com/whisperfish/libsignal-service-rs/issues/206.

Schmiddiii:
After #245 was merged, this is now finally workind and ready for review.

Schmiddiii:
Note that as mentioned in the WF chat, something is wrong with the device name querying the linked devices. It is some base64-encoded string. I don't know how to solve that yet.

gferon:
If it is anything like it used to be for secondary linking, the device name was (at least at some point) encrypted.

Schmiddiii:
Yes, something similar was mentioned in the chat. I may need to investigate this closer before this can be merged.

Schmiddiii:
After https://github.com/whisperfish/libsignal-service-rs/pull/300, it should now be ready.

gferon:
Great addition, thanks!

PR #197, opened 2023-10-09
Attempting to fix quirky contacts sync

gferon:
@Schmiddiii could you try this out?

### DIFF
--- a/presage-cli/src/main.rs
+++ b/presage-cli/src/main.rs
@@ -205,21 +205,11 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn send<C: Store + 'static>(
-    msg: &str,
-    uuid: &Uuid,
     manager: &mut Manager<C, Registered>,
+    uuid: &Uuid,
+    content_body: impl Into<ContentBody>,
+    timestamp: u64,
 ) -> anyhow::Result<()> {
-    let timestamp = std::time::SystemTime::now()
-        .duration_since(UNIX_EPOCH)
-        .expect("Time went backwards")
-        .as_millis() as u64;
-
-    let message = ContentBody::DataMessage(DataMessage {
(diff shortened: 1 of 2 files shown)


presage-cli/src/main.rs
```diff
@@ -654,8 +653,21 @@ async fn run<C: Store + 'static>(subcommand: Cmd, config_store: C) -> anyhow::Re
...
+            use presage::prelude::proto::sync_message;
+
             let mut manager = Manager::load_registered(config_store).await?;
-            manager.request_contacts_sync().await?;
+            let uuid = manager.state().service_ids.aci;
+            let timestamp = std::time::SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("Time went backwards")
+                .as_millis() as u64;
+            let sync_message = SyncMessage {
```
Schmiddiii, line 664: Can we have this in a method in the interface?

presage/src/manager.rs
```diff
@@ -630,43 +630,9 @@ impl<C: Store> Manager<C, Registered> {
...
-            }
-        }
-        Ok(())
-    }
-
     async fn sync_contacts(&mut self) -> Result<(), Error<C::Error>> {
-        let messages = self.receive_messages_stream(true).await?;
-        pin_mut!(messages);
-
+        let _messages = self.receive_messages_stream(true).await?;
```
Schmiddiii, line 634: As you stopped `wait_contacts_sync`, will the `_messages` not drop directly after the method ended and not have a chance to receive any messages? I also may understand futures wrong though.

Schmiddiii:
Does not work for me. It now does not time out anymore, but the contacts are not synced anymore.

gferon:
Most of it was merged and/or being integrated as part of !205

PR #202, opened 2023-11-06
feat: Add receiving mode to the message stream

boxdot:
Add a new method `Manager::receive_messages_with_mode` allowing to
configure how to handle specific sentinel messages. It is supported to
stop the stream after the initial sync (empty Signal message queue) or
after contact sync.


boxdot:
This is partially extracted from #197, but with much smaller change surface.

PR #205, opened 2023-11-15
Split into multiple module, document all structs and general cleanup

boxdot:
There are several goals of this refactoring:

1. Importing of types should be obvious and unique. We try not to
   re-export types (with the exception of the `Manager` and `protos`).
2. The Store trait is split and is not an overarching trait anymore.
   Instead it is combined from several smaller traits, each responsible
   for a non-overlapping feature.
3. All public modules and types are documented (in presage).
4. `Manager` file is split in a file per type state.

Since public exports are removed and reordered, this is a breaking 
change.

TODO: 

- [x] Mark all methods from Manager that are just redirecting to the underlying Store as deprecated for later removal.
- [x] Make all fields of Registered private. Instead expose the relevant data only via getters.
- [x] Remove Serialize/Deserialize from Registered. This would hide the data inside.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,13 +41,13 @@ jobs:
           command: build
           args: --all-targets ${{ matrix.cargo_flags }}
 
-      - name: Run tests
+      - name: Test
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --all-targets ${{ matrix.cargo_flags }}
 
-      - name: Run doc tests
+      - name: Test docs
         uses: actions-rs/cargo@v1
         with:
           command: test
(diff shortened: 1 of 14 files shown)


boxdot:
As the next step, I think we should do the following:

1. Remove all methods from Manager that are just redirecting to the underlying Store. Instead expose a getter for the Store.
2. Make all fields of Registered private. Instead expose the data only via getters.
3. Remove Serialize/Deserialize from Registered. This would hide the data inside.

boxdot:
`SledStore` now implements the following traits
image

btw. I am not entirely sure about the difference between `PrekeysStore` (extracted by me) and the `PreKeyStore`.

gferon:
> btw. I am not entirely sure about the difference between PrekeysStore (extracted by me) and the PreKeyStore.

These are just extra methods we use, on top of what `libsignal` expects to have. We already have `SessionStoreExt` in `libsignal-service-rs` so I'll rename this one `PreKeyStoreExt`.

gferon:
> As the next step, I think we should do the following:

I moved your additional suggestions in TODO and took care of them. Let's ship this 🚀

boxdot:
Thank you for also implementing the TODO.

PR #211, opened 2023-11-19
Upsert a contact when seeing it for the first time

### LINKED ISSUE #98: Unknown contact handling
If i receive a message from an unknown contact it should create a new contact and fetch the profile for the username.

gferon:
Closes #98 and #171

Also fixes updating profile keys (tentatively added a few commits ealier) which was in the wrong place (silly giant `match` statements).


gferon:
Thanks to @nanu-c for the initial inspiration for this PR! (#87)

PR #215, opened 2023-11-27
Update libsignal-service-rs

Schmiddiii:
Fixes <https://github.com/whisperfish/libsignal-service-rs/issues/266> by the upstream PR <https://github.com/whisperfish/libsignal-service-rs/pull/267>.

The fix has not yet been tested by me using Flare.


Schmiddiii:
I tested this with Flare and the fix indeed works as expected.

PR #218, opened 2023-12-10
Fix disappearing messages timer

### LINKED ISSUE #213: Handle Incoming Disappearing Message Timers
Basically an extension of #93. We currently don't handle updates to disappearing message timers which are incoming. This could e.g. in cases where the contact was created from an profile accidentally disable disappearing messages in a chat, or revert a changed disappearing message timer.

We should, for every message that was received, update the stored disappearing message timer to the received value.

Schmiddiii:
Closes #213.

Currently only tested in Note-to-Self.


gferon:
Nice, thanks!

PR #222, opened 2023-12-19
Reconnect websockets lazily on usage

boxdot:
When identified or unidentified websocket is used, it is checked whether it is closed, and in that case it is reconnected. This is less powerful to auto-reconnecting websockets, because the messages stream would not be interrupted. But this is a very simple implementation which works quite well on application level.


gferon:
Nice, thanks!

PR #223, opened 2023-12-19
Drop kyber pre keys on relinking

Schmiddiii:
This caused issues for me with out-of-sync kyber and normal pre keys.


boxdot:
Good catch

boxdot (approved):

Schmiddiii:
Note: Even after the fix and relinking, I am still getting `InvalidPreKeyId`s. So there must also be some other issue somewhere.

gferon:
I can't believe this was missed before, thanks a lot!

PR #225, opened 2023-12-22
Fix for kyber errors

Schmiddiii:
Fixes https://github.com/whisperfish/libsignal-service-rs/issues/273. Note that for the fix to work, the device probably needs to be re-linked.

This was tested for about half an hour yesterday, and continued testing today. No message got lost due to the `InvalidKyberPreKeyId`-error (but I got some other errors which are weird, see the Signal chat; these do only seem to be from one other account and do only seem to be message delivery receipts).

The fix is to register pre-keys (including the kyber pre keys) with the server. This is also what Signal-Desktop does, see [here](https://github.com/signalapp/Signal-Desktop/blob/main/ts/textsecure/AccountManager.ts#L1266).

### DIFF
--- a/presage/src/manager/linking.rs
+++ b/presage/src/manager/linking.rs
@@ -137,11 +137,19 @@ impl<S: Store> Manager<S, Linking> {
                     &registration_data.service_ids
                 );
 
-                Ok(Manager {
+                let mut manager = Manager {
                     rng,
-                    store,
+                    store: store.clone(),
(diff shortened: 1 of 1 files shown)


Schmiddiii:
This fix has now been tested by me and a few others on a beta of Flare for multiple days without any issues. So I will promote the "potential fix" to a "fix".

gferon:
Thanks!

PR #226, opened 2023-12-30
Fix reuse of websocket

Schmiddiii:
When `receive_messages` is called multiple times on the same manager, this could lead to a panic "web socket request handler not in use". This fixes this.

Requires <https://github.com/whisperfish/libsignal-service-rs/pull/274>. Also currently dependent on my fork of lss.


gferon:
Thanks!

PR #230, opened 2024-01-16
Update libsignal-service-rs

Schmiddiii:
~~Note: I am unsure about a few things, which are marked as `TODO`.~~

This should now be ready, with one additional TODO which needs to be implemented once upstream has PNI.

### DIFF
--- a/presage/src/manager/registered.rs
+++ b/presage/src/manager/registered.rs
@@ -23,8 +23,8 @@ use libsignal_service::protocol::SenderCertificate;
 use libsignal_service::protocol::{PrivateKey, PublicKey};
 use libsignal_service::provisioning::generate_registration_id;
 use libsignal_service::push_service::{
-    AccountAttributes, DeviceCapabilities, PushService, ServiceError, ServiceIds, WhoAmIResponse,
-    DEFAULT_DEVICE_ID,
+    AccountAttributes, DeviceCapabilities, PushService, ServiceError, ServiceIdType, ServiceIds,
+    WhoAmIResponse, DEFAULT_DEVICE_ID,
 };
 use libsignal_service::receiver::MessageReceiver;
 use libsignal_service::sender::{AttachmentSpec, AttachmentUploadError};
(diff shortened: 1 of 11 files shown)


presage/src/manager/registered.rs
```diff
@@ -719,6 +720,7 @@ impl<S: Store> Manager<S, Registered> {
             metadata: Metadata {
                 sender: self.state.data.service_ids.aci.into(),
                 sender_device: self.state.device_id(),
+                server_guid: None, // TODO: What value should this have?
```
boxdot, line 723: Signal Desktop has the following comment:

```
  // This should always be set for new messages, but older messages may not have them. We
  //   may not have these for outbound messages, either, as we have not needed them.
```

So, I guess it is ok to leave it empty.

presage/src/manager/registered.rs
```diff
@@ -823,6 +825,7 @@ impl<S: Store> Manager<S, Registered> {
...
```
boxdot, line 828: See above comment

presage/src/manager/registered.rs
```diff
@@ -271,18 +271,19 @@ impl<S: Store> Manager<S, Registered> {
         let (pre_keys_offset_id, next_signed_pre_key_id, next_pq_pre_key_id) = account_manager
             .update_pre_key_bundle(
                 &mut self.store.clone(),
+                ServiceIdType::AccountIdentity, // TODO: What about PNI?
```
boxdot, line 274: This one is tricky. Can't find anything equivalent in Signal Desktop.

boxdot (approved):
Thank you. I think we can remove the todo about the guid, and maybe merge with update pre key bundle for account identity only?

presage/src/manager/registered.rs
boxdot: I just compared to the previous impl of  `update_pre_key_bundle`. It was doing it only for `AccountIdentity` keys. So, at least the behavior is the same.

boxdot:
I tested this commit in Gurk. Somehow I get `failed to send receipt error=libsignal-service sending error: Error decoding JSON response: Invalid padding at line 1 column 296 sender_uuid=e0a46394-2bb1-4264-abc9-469ddbd0c64e`

presage/src/manager/registered.rs
Schmiddiii: From my look at the Signal Desktop code (function `maybeUpdateKeys` of `AccountManager`, both seem to be updated sometimes (in `UpdateKeysListener` and `AccountManager.setPni`). Not sure though if any of those apply to presage though, which only seems to update those after setup, while SD does it every two days as far as I can read.

Schmiddiii:
> I tested this commit in Gurk. Somehow I get `failed to send receipt error=libsignal-service sending error: Error decoding JSON response: Invalid padding at line 1 column 296 sender_uuid=<...>`

Flare does not yet send read receipts, so that did not happen to me. But I guess the error is somewhere in libsignal-service-rs, so maybe ask them about this.

presage/src/manager/registered.rs
rubdos: This is correct. When PNI is fully implemented, pre key generation should also get triggered for the PNI account, but this retains the previous behaviour.

Schmiddiii: I have updated the TODO to change that once it is fully implemented.

gferon:
Quickly tested, LGTM, thanks a lot for working on this!

PR #233, opened 2024-01-30
Update libsignal-service-rs

Schmiddiii:
Main fix is for the error "Invalid padding" for JSON decoding.

See <https://github.com/whisperfish/libsignal-service-rs/issues/288> and <https://github.com/whisperfish/libsignal-service-rs/pull/289>.

### DIFF
--- a/presage/Cargo.toml
+++ b/presage/Cargo.toml
@@ -6,8 +6,8 @@ authors = ["Gabriel Féron <g@leirbag.net>"]
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


Schmiddiii:
Note that I also checked the usage of base64 within presage itself, but presage seems to only use it for storage and not for anything interacting with Signal itself. I therefore don't think it is a problem to keep the presage base64 calls the same, but this could be changed if needed.

gferon:
Thanks!

PR #234, opened 2024-02-15
Add functions to retrieve avatars

Schmiddiii:
A few TODOs are left which I am unsure about:

- [x] When fetching the profile avatar, I am unsure about using the identified or unidentified push service. Note: Unidentified seems to work. I thus chose this.
- [ ] Is there some way of knowing when a profile or profile avatar (or group avatar) gets outdated, to remove that from the cache.
- [x] This may require a migration to remove the already cached profiles, otherwise no avatars for profiles will ever be fetched.
- [x] Is there maybe some better format to use for avatars instead of `Vec<u8>`, e.g. `Bytes`?

### DIFF
--- a/presage-store-sled/src/lib.rs
+++ b/presage-store-sled/src/lib.rs
@@ -496,7 +496,7 @@ impl ContentsStore for SledStore {
     fn group_avatar(
         &self,
         master_key_bytes: GroupMasterKeyBytes,
-    ) -> Result<Option<Vec<u8>>, SledStoreError> {
+    ) -> Result<Option<AvatarBytes>, SledStoreError> {
         self.get(SLED_TREE_GROUP_AVATARS, master_key_bytes)
     }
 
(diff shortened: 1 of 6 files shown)


gferon:
Thanks a lot! We could definitely introduce a convenience wrapper type for avatars if you think it's worth it.

Schmiddiii:
I have now updated this PR. A few notes:

- Using unidentified instead of identified seems to work for profile fetching, I therefore used unidentified (still don't know what is correct though).
- Added the migration mentioned.
- Added `type AvatarBytes = Vec<u8>`.
- Prevented a panic in libsignal-service-rs when fetching a group avatar with empty path.

I am still unsure if there exists some message that notifies when a profile/avatar gets out-of-date.

Schmiddiii:
Note from WF chat: Profile updates seem to be done by WF and SA on a schedule. While presage does not yet have such a schedule, we could certainly do something similar. But I think this can also be done in another MR, I would therefore call this mergeable.

See also:

- <https://gitlab.com/whisperfish/whisperfish/-/blob/main/whisperfish/src/worker/profile_refresh.rs?ref_type=heads>
- <https://github.com/signalapp/Signal-Android/blob/09b9349f6c0cf02688a79d8c2c9edeb8b32dd3cf/app/src/main/java/org/thoughtcrime/securesms/database/RecipientDatabase.kt#L3209>

gferon:
Thanks a lot for your work! I fixed some cosmetic things and some `clippy` warnings, and will merge it now.

PR #237, opened 2024-02-28
chore: fix build

boxdot:
* add missing hex dependency
* use libsignal-service where BytesStream is Send


gferon:
I guess I was a bit drunk when I made the last release 😅

PR #238, opened 2024-03-01
chore: fix build (restore)

boxdot:
* add missing hex dependency
* use libsignal-service where BytesStream is Send


gferon:
Thanks! I'll protect `main` again to avoid my silly unattentive self to break the build again.

PR #240, opened 2024-03-01
chore: modernize CI

boxdot:
* use dtolnay/rust-toolchain
* upgrade actions/checkout to v4
* only install protobuf-compiler


gferon:
I have enabled Dependabot, which sources its info from the Rust Audit database, so you could remove the `cargo-audit` step.

boxdot:
done

PR #253, opened 2024-06-24
upgrade libsignal to v0.51.0

boxdot:


presage/src/manager/registered.rs
```diff
@@ -392,7 +392,7 @@ impl<S: Store> Manager<S, Registered> {
                 .as_secs();
 
             if let Some(expiration) = sender_certificate.and_then(|s| s.expiration().ok()) {
-                expiration >= seconds_since_epoch - 600
+                seconds_since_epoch <= expiration.epoch_millis() + 600
```
Schmiddiii, line 395: I think you will need to convert the `epoch_millis` to seconds here.

boxdot: Good catch. Thank you.

Schmiddiii (approved):

PR #254, opened 2024-07-11
Correctly return the next ID of (signed, kyber and normal) pre-keys

gferon:


presage-cli/src/main.rs
```diff
@@ -704,6 +707,34 @@ async fn run<S: Store>(subcommand: Cmd, config_store: S) -> anyhow::Result<()> {
...
+
+            #[allow(unused)]
+            #[derive(Debug)]
+            struct Stats {
+                aci_next_pre_key_id: u32,
+                aci_next_signed_pre_keys_id: u32,
+                aci_next_kyber_pre_keys_id: u32,
+                pni_next_pre_key_id: u32,
+                pni_next_signed_pre_keys_id: u32,
+                pni_next_kyber_pre_keys_id: u32,
```
boxdot, line 721: I think you could also add the number of current keys

presage-store-sled/src/lib.rs
```diff
@@ -560,6 +569,19 @@ mod tests {
...
 
+    #[quickcheck]
+    fn compare_pre_keys(pre_key_id: u32, next_pre_key_id: u32) {
+        if pre_key_id < next_pre_key_id {
+            assert!(
+                PreKeyId::from(pre_key_id).sled_key() < PreKeyId::from(next_pre_key_id).sled_key()
+            )
+        } else {
+            assert!(
+                PreKeyId::from(pre_key_id).sled_key() > PreKeyId::from(next_pre_key_id).sled_key()
```
boxdot, line 580: missing =

presage-store-sled/src/protocol.rs
```diff
@@ -51,6 +48,38 @@ impl SledProtocolStore<PniSledStore> {
...
+
+    /// Whether to force a pre key refresh.
+    ///
+    /// Check whether we have:
+    /// - 1 signed EC pre key
+    /// - 1 Kyber last resort key
+    async fn needs_pre_key_refresh(&self) -> Result<bool, SignalProtocolError> {
+        let has_signed_pre_keys = self.signed_pre_keys_count().await? > 0;
+        let has_last_resort_kyber_pre_keys = self.kyber_pre_keys_count(true).await? > 0;
+        Ok(has_signed_pre_keys && has_last_resort_kyber_pre_keys)
```
boxdot, line 79: I think it is better to short circuit this query.

boxdot:
Before merging this, there is a problem with V5 migration. Usually it fails and does not store the V5 version, and so tries again next time. With the new V6 migration this will be a problem. Let's fix this first.

I would like to add a test from whisperfish testing the next key id.

presage-store-sled/src/protocol.rs
gferon: what do you mean? if the left hand side is false it'll be short circuited.

gferon:
note: we should only merge this with some sort of backwards compatibility, otherwise we'll make it impossible to decipher any message in the pending message queue. we could find a way to run the migration _after_ we get the empty queue message for example?

presage-store-sled/src/protocol.rs
boxdot: Removed since it is unused

boxdot:
> note: we should only merge this with some sort of backwards compatibility, otherwise we'll make it impossible to decipher any message in the pending message queue. we could find a way to run the migration _after_ we get the empty queue message for example?

I don't see a problem here. We don't actually change any keys, just how they are stored in the storage.

PR #255, opened 2024-07-15
fix: ignore sometimes failing migration v4 -> v5

boxdot:
Without this we are stuck on v4, and always run the migration step. The migration sometimes fails because the there is no properly serialized data.

Also allow to specify the database path and secret at the same time.


gferon:
thanks a lot!

PR #257, opened 2024-07-18
Add add-device and unlink-device to CLI

rlking:


gferon:
Very nice, thanks a lot! I was just thinking of a small improvement: (in another PR) do you think you could move these new commands under a `device` subcommand `device`?

presage-cli/src/main.rs
```diff
@@ -90,6 +90,22 @@ enum Cmd {
...
+    #[clap(
+        about = "Add a new secondary device to this (primary) device via URL (see link-device)"
+    )]
+    AddDevice {
+        #[clap(
+            long,
+            short = 'u',
+            help = "The URL (that is represented as QR code) created by the secondary device (see link-device)"
+        )]
+        uri: String,
```
gferon, line 102: you should be able to use the `Uri` type directly IIRC

rlking: I'll check, I also need to force push since my rebase or squash somehow added those old commits (which are already on main)

rlking: Thx for the hint, I changed it to url directly

rlking:
> Very nice, thanks a lot! I was just thinking of a small improvement: (in another PR) do you think you could move these new commands under a `device` subcommand `device`?

I'll have a look next time, I'm still missing the unregister command. Will checkout the subcommands then

gferon:
Thanks!

PR #260, opened 2024-07-29
feat: upgrade libsignal-service incl. message editing

boxdot:
https://github.com/whisperfish/libsignal-service-rs/pull/312


gferon:
Geil, thanks!

PR #268, opened 2024-09-15
Add support for versioned expire timers

GranPC:
See whisperfish/libsignal-service-rs#321 for more details

NB: my Rust skills compare unfavorably to those of an amoeba, and as such it is entirely possible I have done something horribly wrong.

### DIFF
--- a/presage/Cargo.toml
+++ b/presage/Cargo.toml
@@ -7,8 +7,8 @@ edition = "2021"
 license = "AGPL-3.0-only"
 
 [dependencies]
-libsignal-service = { git = "https://github.com/whisperfish/libsignal-service-rs", rev = "5470ce6a2" }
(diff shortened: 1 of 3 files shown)


Schmiddiii:
@GranPC Seems like you forgot to update the `Cargo.toml` to point to the newest commit of `libsignal-service-rs`, and rebuild `presage` (to update `Cargo.lock`).

gferon:
Thanks a lot for the PR! I introduced a function to make sure the logic is shared when sending both direct and group messages. I also realized there are a bunch of other issues unrelated to your changes: I'm pretty sure we can add a field to the `groups_v2::Timer` struct of `libsignal-service-rs` since it has most likely been introduced in the Protobuf messages.

PR #273, opened 2024-10-16
fix: remove CacheCell in favor of OnceLock

RisaI:
In this PR, the `CacheCell` is entirely removed in favor of `std::sync::OnceLock`. This struct was used previously only in the `Registered` manager state and semantically it was equivalent to `OnceCell`. However, this made the struct `!Sync`, unlike all the other `Manager` states. This seems like an oversight and forces the user to wrap the manager in a `Mutex` to make it `Sync` (such case was a motivation for this change).


gferon:
Thanks a lot, this is definitely a welcome improvement! Some background: [`CacheCell` is nearly 3 years old](https://github.com/whisperfish/presage/commit/6c24c0cd597200dd3796c968d1d5a675af041088) and Rust 1.70 is a bit fresher 😄

PR #277, opened 2024-10-21
Prevent interleaved envelope decryption processes and use the right ServiceCipher

gferon:
Technically, we should only have a set of locks per session (essentially something like a `HashMap<SessionId, Mutex>`) to avoid working on the same sessions concurrently.

### DIFF
--- a/presage/src/manager/registered.rs
+++ b/presage/src/manager/registered.rs
@@ -39,7 +39,7 @@ use rand::rngs::StdRng;
 use rand::SeedableRng;
 use serde::{Deserialize, Serialize};
 use sha2::Digest;
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, Semaphore};
 use tracing::{debug, error, info, trace, warn};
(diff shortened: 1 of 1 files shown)


presage/src/manager/registered.rs
```diff
@@ -633,10 +635,31 @@ impl<S: Store> Manager<S, Registered> {
...
+                                        aci_semaphore.acquire().await.expect("closed semaphore");
+                                    state.service_cipher_aci.open_envelope(envelope).await
+                                }
+                                Some(ServiceId::Pni(_)) => {
+                                    let _permit =
+                                        pni_semaphore.acquire().await.expect("closed semaphore");
+                                    state.service_cipher_pni.open_envelope(envelope).await
+                                }
+
+                                None => todo!(),
```
rubdos, line 659: I'd just log this

rubdos (approved):
LGTM

gferon:
Superseded by #278

PR #287, opened 2024-10-24
Sqlite store impl

gferon:


Schmiddiii:
I would guess `presage-store-sqlite/.sqlx` should be in gitignore?

gferon:
> I would guess `presage-store-sqlite/.sqlx` should be in gitignore?

Not at all, quite the opposite! These are files that you generate with `cargo sqlx prepare` and let you compile check your SQL statements.

Schmiddiii:
Just saying that the [official sqlx repo](https://github.com/launchbadge/sqlx/blob/main/.gitignore#L22) also has it gitignored.
See also the [PR that ignored it](https://github.com/launchbadge/sqlx/pull/3008).

Instead, this should probably be generated and checked in CI.

presage-store-sqlite/presage.sqlite
Schmiddiii, line 1: I think this should also be gitignored. As well as the top-level `presage.sqlite`.

presage-store-sqlite/src/protocol.rs
```diff
@@ -38,7 +55,21 @@ impl SessionStore for SqliteProtocolStore {
         address: &ProtocolAddress,
         record: &SessionRecord,
     ) -> Result<(), ProtocolError> {
-        todo!();
+        let uuid = address.name();
+        let device_id: u32 = address.device_id().into();
+        let record_data = record.serialize()?;
+        query!(
+            "INSERT INTO sessions ( address, device_id, identity, record ) VALUES ( $1, $2, $3, $4 )",
```
Schmiddiii, line 62: As far as I can see, sessions should also be replaced when they already exist (looking at WF).
The same probably also for the various types of keys.

gferon: Can you give it another look when it's not a draft anymore?

I plan on double checking constraints on tables themselves, and most tables already have `PRIMARY KEY(...) ON CONFLICT REPLACE`.

gferon:
superseded by #301

PR #292, opened 2024-11-09
fix: rng state reusage

boxdot:
StdRng was cloned, which created a new rng with the same state. This
means that the same random numbers were generated:

```rust
use rand::{Rng, SeedableRng};
use rand::rngs::StdRng;

let mut rng1 = StdRng::from_entropy();
let mut rng2 = rng1.clone();

assert_eq!(rng1.gen::<u32>(), rng2.gen::<u32>());
```

This commit replaces the `StdRng` with `ThreadRng` in all places where it
was used. It also does not store the rng in manager anymore.

The credits for finding this bug go to @hrdl-github
<31923882+hrdl-github@users.noreply.github.com>.

### DIFF
--- a/presage/src/manager/confirmation.rs
+++ b/presage/src/manager/confirmation.rs
@@ -9,7 +9,7 @@ use libsignal_service::push_service::{
 };
 use libsignal_service::zkgroup::profiles::ProfileKey;
 use libsignal_service::AccountManager;
-use rand::RngCore;
+use rand::{thread_rng, RngCore};
 use tracing::trace;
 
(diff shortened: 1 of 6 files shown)


gferon:
Thanks! Linking https://github.com/whisperfish/libsignal-service-rs/pull/342 here since this is also related (if somebody wants to look into it later).

gferon (approved):

gferon:
Adding a mention of https://github.com/whisperfish/presage/commit/9f717c73f582d5cb79b30090c6b21be86aece17f where the mistake was first introduced.

PR #293, opened 2024-11-10
Expose QueueEmpty and Contacts

### LINKED ISSUE #231: Propagate `QueueEmpty` signal
This prevents a few nice-to-have features of clients, e.g. notifying only messages which have not yet read or needlessly reordering the channels a lot when there are many messages to receive.

Note that the `InitialSync` is not really a perfect solution for that, e.g. if the device has been running for a while but was disconnected from the internet, `InitialSync` probably will not really help.

gferon:
Modify the API to remove `ReceivingMode` in favor of being able to keep the receiving messages loop running (= websocket open) while still handling the different phases in client code. There's an example in `fn send` in `presage-cli` on how to wait for all initial messages (until `QueueEmpty` arrives) and then start sending messages.

This will help with implementing https://github.com/boxdot/gurk-rs/issues/196

Closes #231


presage-cli/src/main.rs
```diff
@@ -789,9 +826,16 @@ async fn run<S: Store>(subcommand: Cmd, config_store: S) -> anyhow::Result<()> {
                 println!("{contact:#?}");
             }
         }
-        Cmd::RequestContactsSync => {
+        Cmd::SyncContacts => {
             let mut manager = Manager::load_registered(config_store).await?;
-            manager.sync_contacts().await?;
+            let sync_message = SyncMessage {
```
Schmiddiii, line 831: Is there a reason that this does not use `request_contacts`?

Schmiddiii (approved):
Did not yet implement it into Flare, just took a look at the code.

presage/src/manager/mod.rs
```diff
@@ -9,7 +9,7 @@ use std::fmt;
 
 pub use self::confirmation::Confirmation;
 pub use self::linking::Linking;
-pub use self::registered::{ReceivingMode, Registered, RegistrationData, RegistrationType};
+pub use self::registered::{Received, Registered, RegistrationData, RegistrationType};
```
boxdot, line 12: Maybe it makes send to export it from the location where it is actually defined, that is, from `crate::model::messages::Received`?

presage/src/model/messages.rs
```diff
@@ -0,0 +1,13 @@
...
+#[derive(Debug)]
+pub enum Received {
+    /// when the receive loop is empty, happens when opening the websocket for the first time
+    /// once you're done synchronizing all pending messages for this registered client.
+    QueueEmpty,
+
+    // got contacts synchronization
+    Contacts,
+
+    Content(Content),
```
boxdot, line 12: Content is 4 KiB in memory. That's why clippy complains about huge size difference on the variants. I think it makes sense to box it.

presage/src/model/messages.rs
```diff
@@ -0,0 +1,13 @@
+use libsignal_service::prelude::Content;
+
+#[derive(Debug)]
+pub enum Received {
+    /// when the receive loop is empty, happens when opening the websocket for the first time
+    /// once you're done synchronizing all pending messages for this registered client.
+    QueueEmpty,
+
+    // got contacts synchronization
```
boxdot, line 9: Maybe add a comment for new users where to get contacts after they got this message?

boxdot (dismissed):

PR #310, opened 2025-03-26
Add tokio::rt feature flag

gferon:
To be able to use [block_in_place](https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html).

This is unfortunately a leaky abstraction from tokio. I would however not expect any user of this library to use anything but the default tokio rt-multi-thread runtime.

Fixes standalone compilation of `presage` and docs generation in CI.


boxdot (approved):
Thanks for fixing it.

PR #312, opened 2025-04-03
fix(store-sqlite): locked db errors

boxdot:
While holding a transaction, we were using another connection from the pool.


gferon:
Nice catch!

PR #314, opened 2025-04-26
Offload large attachment decryption to another thread

kaspar030:
#307 wrapped `decrypt_in_place()` in `block_in_place()`, which makes it impossible to run the manager on a single-threaded executor:

```
thread '<unnamed>' panicked at /home/kaspar/.cargo/git/checkouts/presage-ce211e77a0d9397d/0da7b57/presage/src/manager/registered.rs:1045:9:
can call blocking only when running on the multi-threaded runtime
```

I guess there was an issue with larger attachments. So this PR makes tokio spawn a thread for the decryption, if the attachment size exceeds a threshold value. I arbitrarily set that threshold to 100k.

### DIFF
--- a/presage/src/manager/registered.rs
+++ b/presage/src/manager/registered.rs
@@ -38,7 +38,7 @@ use rand::rngs::ThreadRng;
 use rand::thread_rng;
 use serde::{Deserialize, Serialize};
 use sha2::Digest;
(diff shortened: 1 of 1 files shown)


gferon:
Awesome, thanks a bunch!

PR #316, opened 2025-05-12
fix: cache sender certificate

boxdot:
The sender certificate field in `Registered` was not updated in all instances of the manager after the certificate was fetched. Now the field is cached behind a reference-counted mutex. Additionally, we store the certificate in the store, and it is loaded on startup.

Also, make the state field in the manager immutable. This makes it impossible to modify the state by value.


gferon:
Thanks a lot! This was definitely is a great fix.

PR #321, opened 2025-06-13
Fixes for the sqlite store

Schmiddiii:
Those were required to port Flare to the sqlite store. It is just two changes:

- Make the iterators of the store `Send + Sync` (Flare has a separate thread for presage and the store, and will need to send this iterator to the UI thread).
- Add the `runtime-tokio` feature to sqlx. Otherwise sqlx crashed for me during startup. I could probably also add that feature in the Flare dependencies if you don't want to hardcode tokio in there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3223,6 +3223,8 @@ dependencies = [
  "smallvec",
  "sqlformat",
  "thiserror 1.0.65",
+ "tokio",
+ "tokio-stream",
  "tracing",
  "url",
  "uuid",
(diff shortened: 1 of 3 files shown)


Schmiddiii:
Also note that (as discussed in Signal) the order of the message iterator changed, where the sled store ordered messages by increasing timestamp (with the option to reverse it, as the returned iterator was double ended), sqlite now directly returns the messages ordered decrementally by the timetamp. Can either be fixed in presage or Flare, whereever you prefer.

gferon (approved):

gferon:
Thanks!

PR #329, opened 2025-06-28
Update deps

the-ssd:
Update Cargo.lock
Builds locally on my machine


gferon:
Thanks! We only use the lockfile for CI (since this crate is a library) but if it works, it works!

the-ssd:
> Thanks! We only use the lockfile for CI (since this crate is a library) but if it works, it works!

I guess it doesn't work in CI, sorry

PR #337, opened 2025-10-17
Revert "Force libsqlite3-sys with sqlcipher support"

gferon:
Reverts whisperfish/presage#335 until we add a piece of code that encrypts existing databases that were not encrypted before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,18 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


hoehermann:
This sounds like encryption – when implemented again – will be mandatory. When stored on my own machine, I do not want my personal information to be obfuscated in any way. Is there a setting to opt-out?

PR #338, opened 2025-10-17
Update libsignal-service-rs

### LINKED ISSUE #333: Unable to link device with presage-cli from commit 9e09db0
Testing with HEAD on 2025-09-08 on macOS 15.6.1 ARM, I could not use the QR code (on the phone with Signal installed) or the URL (on desktop with Signal installed) to link persage-cli to an existing account. 

The version of gurk from homebrew could link. Examining the gurk source suggested that presage commit d3a760088 was used.

A clean build and removing the database was able to link.

Git bisect suggests the first bad commit is 9e09db03a

```
% git bisect log
git bisect start
# status: waiting for both good and bad commits
# good: [d3a760088617b657b41ccd4f43e4629e3c3188ac] fix: cache sender certificate and make manager state immutable (#316)
git bisect good d3a760088617b657b41ccd4f43e4629e3c3188ac
# status: waiting for bad commit, 1 good commit known
# bad: [9e09db03af163f0f054aa033ea2cc1731f46c819] chore: upgrade to libsignal 0.77.1 (#319)
git bisect bad 9e09db03af163f0f054aa033ea2cc1731f46c819
# good: [65fbbeea959e9cda4f3b440499c0ae0779b0a354] feat: ensure Manager is Send (#323)
git bisect good 65fbbeea959e9cda4f3b440499c0ae0779b0a354
# good: [adbfc0aeb98082450dc3209a0002e9ff7e71eae3] Deprecate sled store (#331)
git bisect good adbfc0aeb98082450dc3209a0002e9ff7e71eae3
# first bad commit: [9e09db03af163f0f054aa033ea2cc1731f46c819] chore: upgrade to libsignal 0.77.1 (#319)
```

Schmiddiii:
This adds new parameters to mark_kyber_pre_key_used of the stores. ~I have no idea what those parameters are doing, and am just ignoring them for now. If you know whether they are needed and what to do about them, please tell me.~

~This also has a few deprecations in -cipher, I've ignored those for now.~

Fixes #333.


Schmiddiii:
"Fixed" the warnings by allowing deprecations in `presage-store-cipher`, as I think a proper fix for those deprecations is not possible from our side as the deprecated function is from a dependency of `chacha20poly1305` that we cannot just update.

As `presage-store-cipher` is also only used by `presage-store-sled`, that is deprecated since #331, I've decided to also deprecate `presage-store-cipher`. Feel free to undo this deprecation if you disagree.

Schmiddiii:
@gferon Could you take a look at whether what I'm doing here is correct? I'd like to do a release of Flare fixing linking sooner rather than later.

Schmiddiii:
I already have code which should in theory work, but I'll add a few tests just to make sure. I'll push the code soon. But thanks for the offer.

Schmiddiii:
I forgot that we probably want a transaction around the entire function. I'll check how to do that with sqlx.

PR #342, opened 2025-11-16
Add --stop-after-empty-queue flag to presage-cli sync

thandal:
Break when receive hits QueueEmpty


hoehermann:
This makes it impossible to print new messages as they come in.

kurotych:
+1 to having this cli argument

Issue #346, opened 2025-11-24
Ship `.sqlx` as part of `presage-store-sqlite` package


jcgruenhage:
`.sqlx` is not available when using `presage-store-sqlite` from vendored sources, requiring workarounds for fetching it separately. For examples, see these:
- https://github.com/flathub/de.schmidhuberj.Flare/blob/09faecd07f1b2069c993e38fb50662768947c490/de.schmidhuberj.Flare.json#L126-L137
- https://github.com/NixOS/nixpkgs/blob/050e09e091117c3d7328c7b2b7b577492c43c134/pkgs/by-name/fl/flare-signal/package.nix#L56-L69
- https://github.com/jcgruenhage/cports/blob/e9ab7ae188e9e05f7b8b90dd84b3868f910db0c8/user/gurk/template.py#L16-L28

jcgruenhage:
Oh, my mistake then. I thought we were fetching presage from crates.io when vendoring, but apparently this assumption was wrong.

It also affects gurk if built in an environment without network access, like Chimera Linux, where I ran into it while packaging gurk.

sqlx supports it at the workspace and manifest level, you can even override it if necessary. I've tested it locally, just works after moving the directory. Do you want me to open a PR, or do you want to do it yourself? Your comment sounds like you're generally in favor of having `.sqlx` inside `presage-store-sqlite`.

PR #347, opened 2025-11-28
Move sqlx query cache into presage-store-sqlite

jcgruenhage:
Discussion in #346.


Schmiddiii:
> If the database URL is relative to your working directory and absolute, yes then it'll fail. 

Yes, absolute paths indeed do work, but relative ones don't. Still feels like a bug if I'm in the `presage-store-sqlite` directory and export `DATABASE_URL=sqlite:db.sqlite` that it requires a `db.sqlite` in both the root of the repo and in `presage-store-sqlite` to run `cargo sqlx prepare`.

> Right, I had overlooked the just file, sorry. Will make a follow up PR fixing this

Also found that after the PR was merged (I have never used this file yet). Already talked with @gferon about it and he mentioned he will update it. But feel free to open an PR if you are faster.

PR #352, opened 2025-12-29
Fix primary device mode

### LINKED ISSUE #328: Signal destop spinning endlessly trying to sync contacts and groups with linked to presage
Not 100% sure about the background and where it should be really fixed, but this issue is imho serious enough to have it easily visible.

starting Signal desktop 7.52.0 and try to link it to presage with the url given in the qr code: 
```
$ ./presage-cli add-device --url 'sgnl://linkdevice?uuid=xxxxxxxxxxxxxxx%3D%3D&pub_key=xxx%2Bxxx%2Bxxx&capabilities=backup3'
Added new secondary device
```

QR-Code goes away and Signal Desktop says "Syncing contacts and groups", nothing happens after a long time.

./presage-cli does not show "Signal Desktop" in list-device despite the message "Added new secondary device." So it is lying.

Schmiddiii:
This attempts to fix all things required to allow using presage as a primary device.

This fixes #328.

TODO:

- ~~Migration for deleting the `device_id` column for `identites`.~~
- ~~Answer some more sync requests (does not seem to be strictly needed, but are implemented in Whisperfish):~~
    - ~~Keys~~
    - ~~Blocked~~
    - Configuration: We don't have any way to configure that yet.
    - ~~Contacts (real implementation, no dummy).~~
 - ~~Figure out why listing devices does not work.~~


Schmiddiii:
This PR now pretty much depends on #343 as presage is not yet up-to-date with the libsignal-service-rs main branch, and as this PR now depends upon https://github.com/whisperfish/libsignal-service-rs/pull/383 , which still needs to be updated to support the main branch in order to be merged.

Schmiddiii:
Note: The migration present here is actually slightly wrong, and can fail in the case that there are different records with the same address and record but different device ID. See [this downstream bug](https://gitlab.com/schmiddi-on-mobile/flare/-/issues/286). Sadly, I already shipped the migration to Flare users (I had a beta release with this primary device mode, but then came the hotfix release a few days ago). So I'm not sure if we can fix the migration properly. Or do you have any ideas?

boxdot:
> Note: The migration present here is actually slightly wrong, and can fail in the case that there are different records with the same address and record but different device ID. See [this downstream bug](https://gitlab.com/schmiddi-on-mobile/flare/-/issues/286). Sadly, I already shipped the migration to Flare users (I had a beta release with this primary device mode, but then came the hotfix release a few days ago). So I'm not sure if we can fix the migration properly. Or do you have any ideas?

There is always a way to add another migration, to iterate over all migrations in Rust and skip the wrong migration. If it was already applied, then ok. But next migration must be compatible with it though. If it was not applied, then the next migration can be just applied.

gferon:
Amazing, thanks!

Issue #356, opened 2026-01-09
Add a method to help send link previews


gferon:
Finally ended up filing the issue I wanted to offer to @nezmaka for his first contribution 😄 

The `Manager<Registered>::send_message` method accepts anything that can be turned into a `ContentBody` (from `libsignal-service-rs`). This means you can build anything that official Signal clients accept as message content and send it. While this might be sufficient, there are some cases where extra types or helper methods might be beneficial, and preview HTTP(s) links might be one of them.

I would like to add the approach that Signal-Android takes, which can be summarized for `presage` to:
- Add some functionality to build the preview using HTML5 OG tags:
  - Use `reqwest` to fetch the page in the link, parse the HTML content if applicable
  - Check mimetype of response?
  - Use something like `html5ever` or `scraper` (from the top of my head) to parse the og:tags in the `<head>`
  - Populate the struct with what's interesting to us
- Provide a conversion to the `ContentBody` type in the way of your choice, the issue being that you might actually want to also do other things to this message

On the protobuf side of things, you can add as many link previews as you want to a `DataMessage`, which makes me think we should add a trait that operates on this struct and adds the data. Something simple like:

```rust
pub trait WebLinkPreview {
    async fn add_link_preview<T: IntoUrl>(&mut self, url: T);

    fn remove_link_preview<T: IntoUrl>(&mut self, url: T);
}

impl WebLinkPreview for DataMessage {
}
```

Some references:
- [protobuf field preview inside of DataMessage](https://github.com/whisperfish/libsignal-service-rs/blob/main/protobuf/SignalService.proto#L323)
- [protobuf definition of Preview](https://github.com/whisperfish/libsignal-service-rs/blob/main/protobuf/SignalService.proto#L374)

Schmiddiii:
> I noticed that the display of the Preview is not implemented on the receiving side. Would it make sense to add it at the same time? (and if so, should it go to libsignal-service-rs's content.rs or presage-cli's main.rs?)

You can add it to `presage-cli` if you want, could be useful for debugging. `libsignal-servise-rs` would not be the place for something like that, as each application should choose itself how to display it (i.e. it will likely look different in a TUI vs GUI app).

> I'm also wondering to what extent the project follows the official Signal application.

[...]

PR #358, opened 2026-01-20
Added display of data_message.preview

nezmaka:
I updated the `format_data_message` function to display the `DataMessage`'s previews. 
- I made some assumptions about how the previews would best match the current formatting of incoming messages. Currently each preview is displayed on a new line. All fields are printed but `image`, as I was not sure how to best render the `AttachmentPointer`.
- Since now the printed message consists of a `body` and potentially `preview`, I thought it could be a good idea to wrap the former, for example with a `==BEGIN MESSAGE==` and `==BEGIN MESSAGE==` header and tailer, respectively, to differentiate a message that has something that looks like a preview display at the end of its body, from a message that really has a preview, but that seemed like a pretty big change.

Related to https://github.com/whisperfish/presage/issues/356

### DIFF
--- a/presage-cli/src/main.rs
+++ b/presage-cli/src/main.rs
@@ -381,16 +381,20 @@ async fn print_message<S: Store>(
         if !data_message.preview.is_empty() {
             preview_appendix = "\nLink previews:".to_string();
             for (i, preview) in data_message.preview.iter().enumerate() {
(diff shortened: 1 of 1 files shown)


Schmiddiii (commented):
Please run `cargo fmt` to format the code, and `cargo clippy` to show some more improbements to the code.
Otherwise, looks good to me.

PR #359, opened 2026-01-27
feat: Add url previews to outgoing messages

nezmaka:
The implementation is largely finished, however:
- I have not yet implemented the functionality for `send_message_to_group`.
- While a print statement before the `sender.send_message` block shows that the message contains a populated Preview vector, when sending a message via Presage to either the Linux desktop client or the Android mobile client, no preview is displayed. Since I couldn't send and receive with Presage at the same time (maybe I can spin instances on different sockets in parallel?) I don't know whether the `preview` gets dropped somewhere before being sent or something happens to it when it is received.
- My branch requires a few crates (mime, regex, reqwest, scraper). I suppose these are mostly their dependencies, but `Cargo.lock` has gotten quite a few new lines, and I took the conservative step and left it out of the commit. I felt I should be cautious about it. 

A few questions:
- I read only the parts of the code which were relevant to the task at hand. During this journey I couldn't really divine what the separation between `manager` and `model` was based on. I made a guess that the `previews` should go to `model`, but I don't expect to have gotten it right. I'll be very happy to know where my code fits best and how to best import it.
- I left handling the content of `preview.image`, a `<AttachmentPointer>` to the end. I looked briefly for existing code that could suggest how to build it. I could look further, but if somebody could give me a tip or a pointer as to how to create an <AttachmentPointer> based off a url to an image, it would make my job easier.
- Can anybody tell me what the preview's date field (a `u64`) is for? It is not part of the [the Open Graph protocol](https://ogp.me/)'s specification, and none of the messages that I sent with the official Signal apps had a preview that contained a non-None/ 0 value.

https://github.com/whisperfish/presage/issues/356

### DIFF
--- a/presage/src/model/previews.rs
+++ b/presage/src/model/previews.rs
@@ -1,0 +1,434 @@
+use futures::stream::{self, StreamExt};
+use libsignal_service::sender::AttachmentSpec;
+use mime::Mime;
+use regex::Regex;
+use reqwest::header::CONTENT_TYPE;
+use reqwest::IntoUrl;
+use scraper::{Html, Selector};
+use std::str::FromStr;
+use url::Url;
+
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum Error {
+    #[error("Reqwest error: {0}")]
+    ReqwestError(#[from] reqwest::Error),
+    #[error("The response does not contain a Content-Type header")]
+    MissingContentTypeHeadersError,
+    //#[error("Err error: {0}")]
+    //ConversionError,
+    #[error("Err error: {0}")]
+    UnsupportedMimeTypeError(String),
+    #[error("Downloaded file does not look like an image")]
+    NonImageBytesError,
+    #[error("ToStr error: {0}")]
+    ToStrtError(#[from] reqwest::header::ToStrError),
+    #[error("FromStr error: {0}")]
+    FromStrError(#[from] mime::FromStrError),
+}
+
+#[derive(Clone, PartialEq, Debug)]
+pub struct PreviewContent {
+    pub url: Option<String>,
+    pub title: Option<String>,
+    pub image: Option<Vec<u8>>,
+    /// The MIME (e.g. `image/jpeg`) of `image`, detected from the response's
+    /// `Content-Type` header or, failing that, sniffed from the image bytes themselves:
+    pub image_content_type: Option<String>,
+    pub description: Option<String>,
+    pub date: Option<u64>,
+}
+
+pub async fn generate_preview_from_url<T: IntoUrl>(
+    url: T,
+    client: &reqwest::Client,
+) -> Result<PreviewContent, Error> {
+    let url = url.into_url()?;
+    // a default url value in case the og:url tag is missing
+    let default_url_preview = url.clone().host_str().map(|y| y.to_string());
+
+    let mut preview: PreviewContent = {
+        let mut headers = reqwest::header::HeaderMap::new();
+        headers.insert("user-agent", "WhatsApp/2".parse().unwrap());
+
+        let response = client
+            .get(url)
+            .headers(headers)
+            .send()
+            .await?
+            .error_for_status()?;
+
+        let headers = &response.headers();
+
+        match headers.get(CONTENT_TYPE) {
+            None => return Err(Error::MissingContentTypeHeadersError),
+            Some(content_type) => {
+                let content_type = content_type.to_str()?;
+                let content_type = Mime::from_str(content_type)?;
+                match (content_type.type_(), content_type.subtype()) {
+                    (mime::TEXT, mime::HTML) => {
+                        let html = response.text().await?;
+                        generate_preview_from_html(&html, client).await
+                    }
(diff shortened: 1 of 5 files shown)


presage/src/model/previews.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+    urls
+}
+
+async fn fetch_html_of_url<T: IntoUrl>(url: T) -> Result<String, String> {
+    let mut headers = reqwest::header::HeaderMap::new();
+
+    // appeared in tutorials for scraping, but seems not to be necessary:
+    // headers.insert("authorization", "<authorization>".parse().unwrap());
+    headers.insert("user-agent", "CUSTOM_NAME/1.0".parse().unwrap());
```
Schmiddiii: Signal Android and Signal Desktop pretend to be WhatsApp here (i.e. user agent `WhatsApp/2`). I wonder if we want to do the same.

presage/src/model/previews.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+    urls
+}
+
+async fn fetch_html_of_url<T: IntoUrl>(url: T) -> Result<String, String> {
+    let mut headers = reqwest::header::HeaderMap::new();
+
+    // appeared in tutorials for scraping, but seems not to be necessary:
+    // headers.insert("authorization", "<authorization>".parse().unwrap());
```
Schmiddiii: This is not needed for our use-case. This is only needed if you want to scrape a site that is only accessible behind a login page.

Schmiddiii (commented):
(forgot about this PR the last few days, sorry)

presage/src/model/previews.rs
nezmaka: ooh it's getting political ;)

presage/src/model/previews.rs
```diff
@@ -0,0 +1,227 @@
...
+    let mut _previews: Vec<Preview> = Vec::new();
+    let mut preview = Preview {
+        url: None,
+        title: None,
+        image: None,
+        description: None,
+        date: None,
+    };
+
+    let re = Regex::new(r"^og:(.*)").unwrap();
```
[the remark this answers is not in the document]
nezmaka: I'll look into it, thank you.

nezmaka:
Is there a way to send and receive in parallel using presage-cli?

nezmaka:
> You'll first need to get the image data using [reqwest](https://docs.rs/reqwest/latest/reqwest/), similar to what you have already done with the HTML. You can then use the `Manager::upload_attachment` function to upload the file you got and get back an `AttachmentPointer`.

Is the `AttachmentSpec`'s `file_path` crucial for the `AttachmentPointer`? From what I'm getting, within the current usage, when attaching a file to a message, the file is read from disc and its path is retrieved, which is then added to the `AttachementSpec`. 
I left it tentatively empty (None) in my implementation of previewing, but perhaps I should explicitly save the preview image to disc or —is it automatically saved into a temp dir?— get its path to attach it? There's probably another reason why the one preview that was generated successfully did not contain an image (which was generated when sending the same url with one of the official clients) but I'm asking just in case.

Schmiddiii:
> currently sending some urls with Presage generates a preview (albeit without an image) but most do not

I would recommend adding some `dbg!(...)` macros around your code, then you can see whether the preview is actually set by presage and Signal ignores it, or that maybe you have some logic mistake somewhere with certain URLs.

> Is the AttachmentSpec's file_path crucial for the AttachmentPointer?

I don't know that. I would recommend sending a message containing a preview from Signal to presage to check what Signal does. Anyway, I don't think it is required to save that attachment somewhere to the file system, you can just set the string directly.

nezmaka:
> I would recommend adding some `dbg!(...)` macros around your code, then you can see whether the preview is actually set by presage and Signal ignores it, or that maybe you have some logic mistake somewhere with certain URLs.
Great tip, thank you! I had since figured out that one issue could be that the implementation for grabbing that image wars missing/ commented out :)

nezmaka:
There are a few issues with my current implementation. One I'd like to ask about is the following. When using the cli to send a message to myself, I can see that the message generated by presage contains the preview as far down as the `registered::Manager::send_message` function

https://github.com/whisperfish/presage/blob/b575f2e83932e9d8bb29d281bc2ae2c93d9758f4/presage/src/manager/registered.rs#L931C1-L940C21

That is, the `content_body` passed to `libsignal`'s  `sender.send_message` function contains the preview, yet on the Signal official clients no such preview is displayed, nor, therefore, supposedly, is it received. 
I'm not sure how to further debug this. Any clue?

Schmiddiii:
Maybe try receiving a message with a preview from a Signal device with presage and compare the `ContentBody` you receive field-by-field with what you are sending. Other than that, I would also know of no way how this can be debugged.

Also note that you until now only implemented receipts for direct messages, not for groups as far as I can see. So make sure you're not actually testing this in a group.

Schmiddiii:
You can do it as separate CLI invokations, I think. First receive messages while sending a preview with Signal Android, then another invokation trying to send a message with a preview.

PR #363, opened 2026-02-15
Fix refresh registration

Schmiddiii:
Previously, the registration refresh future could get cancelled when a message is received by the incoming messages loop. This commit fixes this by awaiting for the registration refresh future after a message was received.


gferon:
Definitely an oversight out of lack of sleep, this makes sense and works as it was originally intended, thanks a lot!

I merged the original change with this issue because I was only testing this after linking when the message queue was fresh...

PR #364, opened 2026-02-15
Lock message sending until the queue is empty

Schmiddiii:
Required due to some cryptography things I don't really understand. Not sure about whether the implementation as done here makes sense, or if a mutex or something else would be better to use.

### DIFF
--- a/presage-cli/src/main.rs
+++ b/presage-cli/src/main.rs
@@ -131,6 +131,12 @@ enum Cmd {
     Sync {
         #[clap(long = "notifications", short = 'n')]
         notifications: bool,
+        #[clap(
+            long,
+            short = 'q',
+            help = "Exit after processing the messages in the queue (similar to the syncing of Signal Desktop)"
+        )]
+        stop_after_empty_queue: bool,
     },
     #[clap(about = "List groups")]
     ListGroups {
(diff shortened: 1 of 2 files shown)


presage/src/manager/registered.rs
```diff
@@ -183,6 +189,15 @@ impl<S: Store> Manager<S, Registered> {
         })
     }
 
+    async fn wait_until_allowed_to_send(&self) {
+        self.state
+            .send_lock_receiver
+            .clone()
```
gferon: you can actually only store the Sender, and call `subscribe()` to get a fresh Receiver on which you can call `wait_for`.

presage/src/manager/registered.rs
```diff
@@ -636,77 +654,91 @@ impl<S: Store> Manager<S, Registered> {
...
-                                                                length: a.reader.len().try_into().ok(),
+                                        let request = *request;
+                                        let store = state.store.clone();
+                                        let send_lock = state.send_lock.clone();
+                                        let mut message_sender = state.message_sender.clone();
+                                        let service_ids = state.service_ids.clone();
+
+                                        // In case such requests are received in the initial sync, they can only be answered once the initial sync is finished.
+                                        // Therefore,need to spawn answering to such requests in a new tokio local task.
+                                        tokio::task::spawn_local(async move {
```
gferon: I didn't think we would have to do this :')

PR #367, opened 2026-02-19
Skip instrumenting sqlite data message content body

Schmiddiii:
Lots of noise for no benefit.


Schmiddiii:
Not sure what you mean by having an allowlist or building the span manually. I usually use the `log`  crate instead of `tracing`.

gferon:
Forget about it, it's just a way to do what the macro does, but then the list of fields you log are explicitly listed.

Check this out https://docs.rs/tracing/latest/tracing/span/index.html

At the end, it's two lines of code instead of the macro on top of the function signature.

PR #369, opened 2026-02-22
Integrate CDSI

gferon:
This enables looking up contacts by phone number and username, which lets you start new conversations.

See https://github.com/whisperfish/libsignal-service-rs/pull/398

⚠️ feature disabled by default for the time being, as this builds on macOS only.

This PR also contains the necessary changes to update libsignal-service.

### DIFF
--- a/presage/src/manager/registered.rs
+++ b/presage/src/manager/registered.rs
@@ -8,8 +8,6 @@ use libsignal_service::protocol::{DeviceId, Username};
 use libsignal_service::websocket::account::{
     AccountAttributes, DeviceCapabilities, DeviceInfo, WhoAmIResponse,
 };
-#[cfg(feature = "cdsi")]
-use libsignal_service::{websocket::directory::LookupRequest, protocol::E164};
 use libsignal_service::{
     attachment_cipher::decrypt_in_place,
     cipher,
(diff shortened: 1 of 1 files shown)


gferon:
Merging this since it works (on macOS, because sqlcipher compiles with native crypto there and thus openssl and boring symbols do not conflict). Will fix Linux build ASAP with the patches introduced by @rubdos on rusqlite.

presage/Cargo.toml
```diff
@@ -6,8 +6,12 @@ authors = ["Gabriel Féron <g@leirbag.net>"]
 edition = "2021"
 license = "AGPL-3.0-only"
 
+[features]
+default = [] # not default until we make it compile
+cdsi = ["libsignal-service/cdsi"]
+
 [dependencies]
-libsignal-service = { git = "https://github.com/whisperfish/libsignal-service-rs", rev = "8bbad87b8878b9e4956cdad08d640def94c8a9cd" }
+libsignal-service = { git = "https://github.com/whisperfish/libsignal-service-rs", rev = "c0a3a8802c1cbe0a0e781850c4f6b1b9931b5ee9", features = ["phonenumber"] }
```
Schmiddiii: I think you'll need to add `default-features = false` here to actually disable cdsi.

PR #372, opened 2026-02-27
fix(cli): create sqlite db parent directory automatically

tarasglek:
presage fails to start for me without mkdir -p  of sqlite dir.

This change ensures the sqlite DB parent directory is created automatically before opening the database path.


gferon:
Thanks, this was definitely missing!

tarasglek:
Wow, I've never had someone approve a pr this fast :)

PR #379, opened 2026-03-01
Fix compilation without CDSI

Schmiddiii:
Presage-store-sqlite depended on presage with default features, and therefore CDSI.


Schmiddiii:
I still can't figure it out on my side, and would like to release a new Flare asap.

PR #390, opened 2026-03-30
Add support for updating the account profile

thandal:
Supports setting name, about, emoji, and discoverability.

### DIFF
--- a/presage/src/manager/registered.rs
+++ b/presage/src/manager/registered.rs
@@ -3,6 +3,7 @@ use std::sync::{Arc, OnceLock};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use futures::{future, AsyncReadExt, Stream, StreamExt};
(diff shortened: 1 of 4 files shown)


presage/src/manager/registered.rs
```diff
@@ -563,7 +617,7 @@ impl<S: Store> Manager<S, Registered> {
         // oneshot::channel() or CancellationToken because of !Send constraints in the Store.
         let refresh_registration_task = async move {
             if let Err(error) =
-                set_account_attributes(&mut account_manager, &store_inner, &registration_data_inner)
+                set_account_attributes(&mut account_manager, &store_inner, &registration_data_inner, true)
```
Schmiddiii: This will overwrite `discoverable_by_phone_number=true` every time the message loop is started, making your `set_phone_number_discoverability` pretty much useless.
We'll probably need to store the setting, make the parameter optional, and fetch it when setting account attributes with the parameter not given.

PR #393, opened 2026-04-05
fix: preserve disappearing_messages_timer during group upsert

### LINKED ISSUE #392: upsert_group overwrites disappearing_messages_timer with server-fetched value (always None)
When a group message arrives, `upsert_group()` re-fetches the group metadata
from the Signal server. The server does not store `disappearing_messages_timer`
(it's client-side state managed via `DataMessage.expire_timer` on individual
messages), so the fetched group always has `disappearing_messages_timer: None`.

This overwrites any locally stored timer value that was previously set via
`update_expire_timer()` when processing an incoming message with
`expire_timer` set.

The result: downstream clients calling `store.expire_timer(&thread)` for a
group always get `None`, even if the group has disappearing messages enabled
and the timer was correctly stored moments earlier by `save_message()`.

## Reproduction

1. Enable disappearing messages on a group chat (from another client)
2. Receive a message in that group via presage
3. Observe: `update_expire_timer` correctly stores the timer
4. Observe: `upsert_group` then re-fetches the group and overwrites
   `disappearing_messages_timer` with `None`
5. Subsequent calls to `store.expire_timer(&thread)` return `None`

This can be confirmed with `RUST_LOG=presage=trace` — the trace output
from `upsert_group` shows `disappearing_messages_timer: None` on the
server-fetched group.

## Impact

Downstream clients (gurk-rs, Flare) that rely on `store.expire_timer()`
to set `expire_timer` on outgoing messages will send messages without
the timer, which can disrupt disappearing messages for other participants.

Related: #168 — this bug is a prerequisite for correctly implementing
client-side message expiration.

## Fix

Proposed fix: preserve the locally stored `disappearing_messages_timer`
during group upsert when the server-fetched value is `None`.

## Context

We have a working implementation of disappearing messages in gurk-rs
(boxdot/gurk-rs) that works around this issue by storing the timer
independently on the client side. The implementation does not depend
on this fix but would benefit from it — downstream clients could read
the timer from presage's store rather than maintaining their own copy.

The gurk-rs disappearing messages work is pending upstream review of
prerequisite PRs (boxdot/gurk-rs#519, boxdot/gurk-rs#518).

cwedgwood:
Fixes #392

When a group message arrives, `upsert_group()` re-fetches the group from
the Signal server. The server does not store `disappearing_messages_timer`
(it's client-side state), so the fetched group always has it as `None`.
This overwrites any locally stored timer value.

Fix: read the existing timer before the fetch. After decrypting the new
group, restore the old timer only if the server didn't provide one.
This preserves client-side state while remaining forward-compatible if
the server starts including the timer in future.


gferon:
Thanks a lot, this is a great fix!

Schmiddiii:
Isn't this already fixed by #397?
At least with my tesing of disappearing messages in groups, my fix works as expected.

gferon:
This is true, I literally forgot I merged your patch @Schmiddiii (🧠 has left the room).

cwedgwood:
Closing — this is superseded by #397 which was merged and fixes the same issue. Thanks @Schmiddiii!

PR #407, opened 2026-05-31
Allow non-mutable references for all store functions and some manager functions

Schmiddiii:
This allows parallelizing e.g. fetching avatars. I've kept functions that send messages mut, as those should (to my knowledge) not allow parallel execution.

### DIFF
--- a/presage-cli/src/main.rs
+++ b/presage-cli/src/main.rs
@@ -770,7 +770,7 @@ async fn run<S: Store>(subcommand: Cmd, store: S) -> anyhow::Result<()> {
             uuid,
             mut profile_key,
         } => {
-            let mut manager = load_registered_and_receive(store).await?;
+            let manager = load_registered_and_receive(store).await?;
             if profile_key.is_none() {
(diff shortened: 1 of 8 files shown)


boxdot (approved):
Just be careful with multiple writes on sqlite.

Schmiddiii:
> Just be careful with multiple writes on sqlite.

True, I think it can deadlock? We certainly want to avoid that. As this actually was not that much of a performance improvement than I hoped, I'm also fine with keeping it as-is if that means no deadlocks.

boxdot:
It should not deadlock but it might timeout on write operations. Therefore they should be atomic that is inside a transaction. And you might read stale data if a write transaction is currently running.

Schmiddiii:
Timeouts are still pretty terrible. Given that the store is supposed to be used by clients as well, I don't think risking missuse leading to timeouts is worth the potential performance improvements of fetching avatars in parallel. I'll therefore close it. But if you or @gferon feel different about it, we can still do it.

PR #415, opened 2026-06-20
Fix session-reset

Schmiddiii:
The session-reset mechanism changed a while ago. Implement the new method.


Schmiddiii:
This is very subtly wrong: Deleting and archiving sessions are indeed different. The problem with deleting sessions, as done in here, is that afterwards libsignal_service does not have the information anymore what the other devices are. One therefore needs to archive the session. For that to be handled correctly, https://github.com/whisperfish/libsignal-service-rs/pull/443 is also needed.

Marking as a draft for now.

Schmiddiii:
Temporarily based this MR on #416 as well as on my fork of libsignal-service-rs, to allow implementing the proper fix.

PR #416, opened 2026-06-24
Update libsignal-service-rs

Schmiddiii:
Mostly change of the `Metadata::timestamp` type to `DateTime<Utc>`, as well as the introduction of the new `Metadata::server_timestamp`.


Schmiddiii:
Did not update to the latest commit, as that would require pretty large changes (support for syncing the account entropy pool from the main device, storage of the account entropy pool). Currently too hot where I am to even think about implementing that.

PR #424, opened 2026-07-14
Trying to fix primary registration

### LINKED ISSUE #371: Can't register a device using presage-cli -- "Websocket error: websocket upgrade failed"
With a fresh checkout and fresh build, when going through the captcha / SMS registration process, for example
```sh
presage/target/release$ ./presage-cli --sqlite-db-path /tmp/presage.db.enc --passphrase XXXXXXX register --servers production --phone-number +18XXXXXXX --captcha signalcaptcha://signal-hcaptcha.XXXXX
input confirmation code (followed by RETURN):
XXXXXX
```
it fails with 
```
Error: libsignal-service error: Websocket error: websocket upgrade failed

Caused by:
    Websocket error: websocket upgrade failed
```

Schmiddiii:
Original code by @gferon on the fix-primary-registration branch, adjusted for the current libsignal-service-rs state.

Currently does not work for me on the staging servers, I'm getting an `Unauthorized` error trying to send the captcha code. Looking at WF, the only difference I notice is that they send some MCC/MNC values derived from the phone number carrier, I also tested it but it did not fix it (for my phone number, both were `None`). I don't see any other difference, and WF worked up to a few days ago, which is weird.

Attempts to fix #371.
Related upstream fix: https://github.com/whisperfish/libsignal-service-rs/pull/457


Schmiddiii:
This works now. Turns out that Signal has different captcha-generators for staging and production, and that I copied a trailing `/` from the captcha-generation URL making my captchas invalid. There is not much we can do about the first issue (except documenting, which I did), for the second issue I now implemented parsing the captcha URL to remove unneeded things.

gferon (approved):
Awesome, thanks!

PR #430, opened 2026-08-09
Expose link-time history transfer credentials

emilianscheel:
Adds a backwards-compatible link_secondary_device_with_history entry point while keeping link_secondary_device unchanged. It returns the one-time ephemeral backup key, ACI, device ID, and password required to authenticate Signal's initial transfer_archive flow. Depends on whisperfish/libsignal-service-rs#466; the temporary fork pin will be replaced with the merged upstream revision before this leaves draft.


Schmiddiii:
Thanks for the PR. Any reason we can't just add an optional `ephemeral_backup_key` field to `RegistrationData` and keep the public API otherwise unchanged? Seems to be similar to what Signal Android does (`Database.proto` contains `ephemeralBackupKey` in  `LinkedDeviceInfo`).

